PR #41, opened 2026-04-17
fix(ci): unblock dependabot pipeline + eliminate deprecated Anthropic Temperature

leehopper:
## Summary
- **CI**: Sonar steps now skip gracefully when `SONAR_TOKEN` is unavailable (Dependabot PRs, forks, missing-secret drift). Build+test still runs as smoke check, satisfies branch protection. Job-level `env` so the `if`-condition resolves on the same step.
- **Dependabot**: ignore `eslint@>=10` and `typescript@>=6` until upstream peer-dep ranges (`typescript-eslint`, `@eslint-community/eslint-utils`) widen. Stops PRs #27/#28 from flapping.
- **Anthropic SDK**: eliminate deprecated `MessageCreateParams.Temperature` everywhere (production + eval). Required by SDK 12.16+ which marks it `[Obsolete]` (TreatWarningsAsErrors → build fail). Models released after Claude Opus 4.6 ignore the parameter regardless.
- **PlanConstraintEvaluator**: scope the fast-pace upper-bound check to speed-work types (Tempo/Interval/Repetition). Caught while re-recording: James-injured LongRun legitimately sets fast == easy slower than the calculated easy zone (deliberate conservative coaching for injury recovery).
- **Eval cache**: re-recorded (cache keys changed when Temperature was removed from `ChatOptions`).

## Test plan
- [x] `dotnet build` — 0 warnings, 0 errors
- [x] `dotnet test` (Replay mode) — 417/417 pass
- [x] Pre-push hooks pass (eval-cache-replay green)
- [ ] CI green on this PR (Sonar reports success even without token)

## Follow-ups (separate work, not this PR)
- Merge #35 (frontend minor-and-patch, already passes)
- Close #27 (eslint 10) and #28 (typescript 6) — now ignored by Dependabot
- Rebase #37, #36, #24, #23 to pick up the Sonar fix → merge as they go green

## Summary by CodeRabbit

* **Chores**
  * Updated Dependabot to ignore semver-major updates for selected frontend npm deps
  * Made SonarQube workflow analysis conditional on token availability

* **Configuration**
  * Removed explicit "temperature" setting from coaching model configuration and prompts

* **Tests**
  * Removed and updated tests/assertions related to the removed temperature setting

* **Bug Fixes**
  * Adjusted pace validation to avoid false violations for non-speed workouts

* **Docs**
  * Minor wording update in the POC roadmap referencing temperature settings


leehopper:
Re CodeRabbit's "out of scope" flag on the `PlanConstraintEvaluator` change: it was discovered transitively as part of completing the Anthropic Temperature work, not a separate concern.

**Causal chain:**
1. PR removes `Temperature` from `ChatOptions` in eval tests (required by SDK 12.16+ deprecation).
2. M.E.AI cache hashes `ChatOptions`, so all committed cache fixtures became invalid → must re-record.
3. Re-recording (`./backend/tests/scripts/rerecord-eval-cache.sh`) ran fresh LLM calls and revealed `PlanConstraintEvaluator` flagged James-injured's Saturday LongRun as a violation: `fast_pace 368 > easy_max 358`.
4. Reviewing the response showed the LLM duplicated the easy pace into `fast_pace` (no distinct fast segment), which is a deliberate conservative coaching choice for injury recovery — not an LLM error. The constraint was over-broad: it applied a "fast pace must beat easy max" check to all workout types, including Easy/LongRun/Recovery where there's no fast segment by definition.
5. Without this fix, the re-recorded cache cannot pass `dotnet test` in Replay mode, and the PR cannot merge.

Splitting it into a follow-up would block this PR on a separate eval-fix PR landing first, with no benefit. Inline rationale and a unit-test that still triggers the violation for `Interval` workouts (the case we *do* want to catch) preserves the original intent.

Re docstring coverage (66.67%): warning only, not blocking. Will address in a separate sweep if it becomes an enforced gate.

PR #44, opened 2026-04-17
refactor(training): DEC-042 pure-equation PaceZoneCalculator + DEC-041 value objects

leehopper:
## Summary

DEC-042 pure-equation pace-zone calculator + DEC-041 value objects. Replaces the 56-row transcribed Daniels lookup table (`PaceCalculator.cs`) with a `PaceZoneCalculator` that computes every zone on demand from the 1979 Daniels–Gilbert equations. Renames `VdotCalculator` → `PaceZoneIndexCalculator`. Adds `HeartRateZoneCalculator` with Tanaka max-HR. Lands `Distance`, `Pace`, `PaceRange(Fast, Slow)`, and `TrainingPaces` value objects.

Two mid-flight bugs surfaced and were fixed:
- **Newton-Raphson root equation** was `F·VO₂ = index` instead of `VO₂/F = index` (the Daniels VDOT relation). At index 50 the buggy solver gave a 2:19 marathon and F<R inversion; the correct solver reproduces Daniels' published 3:10:49 M-pace and F-faster-than-R ordering. R-pace tests passed coincidentally because at 3 km `F(t) ≈ 0.997`.
- **`MesoWeekOutput.Days: MesoDaySlotOutput[]`** relied on schema-description text for the 7-day invariant, which constrained decoding doesn't enforce. Priya's `MaxRunDaysPerWeek: 4` triggered the model to hedge with an 8-day array ~67% of re-record attempts. Restructured to seven named `Sunday..Saturday` required properties — invariant becomes structural, `Priya_Constrained_RespectsExactly4RunDays` un-Skip'd and deterministic across three consecutive Record runs.

20 commits, 109 files changed, +2108 / −1016.

### Key additions
- `backend/src/RunCoach.Api/Modules/Training/Computations/` — `DanielsGilbertEquations` (static helper), `PaceZoneCalculator`, `PaceZoneIndexCalculator`, `HeartRateZoneCalculator` and their interfaces.
- `backend/src/RunCoach.Api/Modules/Training/Models/` — `Distance`, `Pace` (value objects), rewritten `PaceRange(Fast, Slow)` and `TrainingPaces`.
- Equation-anchored 56-row `[Theory]` fixture with provenance header citing the 1979 *Oxygen Power* monograph.
- `MesoWeekOutput` restructured to 7 named day properties.
- README trademark disclaimer; `backend/CLAUDE.md` and `backend/REVIEW.md` updated to reflect the shipped rename.

### Key deletions
- `PaceCalculator.cs`, `IPaceCalculator.cs`, `VdotCalculator.cs`, `IVdotCalculator.cs`, `VdotCalculatorTests.cs`, `PaceCalculatorTests.cs` — the old lookup + old class names, fully removed (no forwarding aliases).
- Fully deleted `fix/daniels-pace-table` branch (local + remote) — superseded by the pure-equation approach.

### Post-implementation notes

- Eval cache is unchanged. Test profiles flowed through a `TestPaceCalculator` lookup-table bridge that carried correct Daniels values, so the bug only affected production runtime. The bridge is documented as a DEC-042 follow-up (migrate `TestProfiles` to the real `PaceZoneCalculator` and re-record the cache).
- Two internal VDOT identifiers remain (`FitnessEstimate.EstimatedVdot`, an XML doc comment on `RaceTime`) — internal-only, out of user-facing scope, noted as a rename-on-next-touch follow-up.
- `SmokeTests.GetHealth` is now gated behind `RUNCOACH_INTEGRATION_TESTS=1` so local runs skip in ~0s instead of hanging on WebApplicationFactory + Testcontainers boot for 5 minutes.

## Test plan

- [x] `cd backend && dotnet build` — 0 warnings, 0 errors (`TreatWarningsAsErrors=true`)
- [x] `cd backend && EVAL_CACHE_MODE=Replay dotnet test` — 549 total, 0 failed, 1 skipped (SmokeTest opt-in), 0.2s
- [x] `PaceZoneCalculatorTests.CalculatePaces_FullFixture_AllZonesMatchEquationDerivedValues` — 56×6 assertions pass
- [x] `PaceZoneCalculatorTests.CalculatePaces_Monotonicity_AllZonesOrderedFromSlowToFast` — un-skipped, passes (M > T > I > R > F at index 50)
- [x] `PaceZoneCalculatorTests.CalculatePaces_MPacePrecision_WithinPublishedTableTolerance` — un-skipped, passes (M@50 ≈ 271 s/km, Daniels anchor)
- [x] `PlanGenerationEvalTests.Priya_Constrained_RespectsExactly4RunDays` — un-skipped, deterministic 3/3 Record-mode passes
- [x] `git grep -nE '(VdotCalculator\\b|PaceCalculator\\b|IVdotCalculator|IPaceCalculator|MinPerKm|MaxPerKm|MesoDayOutput)' backend/` — zero matches

## Summary by CodeRabbit

* **New Features**
  * Added heart‑rate zone calculator and a pace‑zone index calculator for deriving fitness and training paces from race performances.

* **Improvements**
  * Rebranded user‑facing “VDOT” to “Pace‑zone index” across UI and messages.
  * Pace outputs now use equation-based calculations with Fast/Slow ranges and an optional Fast‑repetition pace.
  * Weekly plans use explicit Sunday–Saturday day properties for clearer structured output.

* **Bug Fixes**
  * Fixed race‑time solver convergence for more reliable pace predictions.

* **Documentation**
  * Updated trademark disclaimer, roadmap, and supporting docs to reflect changes.


backend/src/RunCoach.Api/Modules/Training/Computations/DanielsGilbertEquations.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// </summary>
+    public static double SolveVelocityForTargetVo2(double targetVo2)
+    {
+        // OxygenCost(v) = targetVo2
+        // OxygenCostQuadratic·v² + OxygenCostLinear·v + (OxygenCostConstant − targetVo2) = 0
+        // Positive root: v = (−b + √(b² − 4·a·c_term)) / (2·a)
+        var cTerm = OxygenCostConstant - targetVo2;
+        var discriminant = (OxygenCostLinear * OxygenCostLinear) - (4.0 * OxygenCostQuadratic * cTerm);
+        return (-OxygenCostLinear + Math.Sqrt(discriminant)) / (2.0 * OxygenCostQuadratic);
+    }
```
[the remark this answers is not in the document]
leehopper: Not resolving — this finding is a false positive.

Your worked example sign-flipped `OxygenCostConstant`. In the actual code, `OxygenCostConstant = -4.60` is negative, so:

- `cTerm = OxygenCostConstant - targetVo2 = -4.60 - 87.57 = -92.17` (not `+92.17`)
- `discriminant = (0.182258)² - 4·0.000104·(-92.17) = 0.033218 + 0.038343 = +0.071561`

The discriminant is positive at the boundary you cited (index 90 / `IntervalFraction = 0.973`), not negative.

More generally: within the valid domain `[MinIndex=25, MaxIndex=90]` × fractions `[0.59, 0.973]`, `targetVo2` ranges from ~14.75 to ~87.57, so `cTerm` is always ≤ -19, so `-4·a·cTerm` is always positive, so the discriminant is always positive. No guard needed — the NaN propagation scenario is not reachable.

Per this project's convention (don't validate for scenarios that can't happen; trust internal guarantees and only validate at system boundaries), no change applied.

ROADMAP.md
```diff
@@ -2,9 +2,9 @@
 
 Living project state. Read this at the start of every session.
 
-## Current Phase: POC 1 Complete — Merging to Main
+## Current Phase: DEC-042 Complete — Pure-Equation Pace-Zone Calculator
 
-POC 1 productionized on `feature/poc1-context-injection-v2`. All POC scaffolding removed (console app, experiment prompts, spec artifacts, poc1 naming). TestProfiles relocated to test project. Eval cache moved from `backend/poc1-eval-cache/` to `backend/tests/eval-cache/`. 290 tests passing in Replay mode (0 failures). Tag `poc1-complete` preserves full POC history.
+POC 1 productionized on `feature/poc1-context-injection-v2` (tag `poc1-complete`). DEC-042 pure-equation pace-zone calculator implemented on `refactor/dec-042-pace-zone-calculator` — all six zones, `VdotCalculator` renamed to `PaceZoneIndexCalculator`, lookup table deleted, eval cache re-recorded.
```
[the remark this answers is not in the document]
leehopper: Partially resolving, keeping the rest intentional.

**Changed (commit 8a652be):**
- Line 127 — softened the stale "all identifiers neutral" claim to match reality
- Line 199 — "VDOT term" → "legacy trademarked term"

**Keeping as-is on lines 7, 57, 122** — these are backticked `VdotCalculator` references to the specific historical class identifier. Per this project's trademark rule in `CLAUDE.md`:

> Internal code identifiers, variable names, private implementation, and historical research artifacts may use VDOT freely until DEC-042's pace-calculator rewrite replaces them with `PaceZoneIndexCalculator` and friends.

The backticked form is a precise reference to the code symbol that existed pre-rename. Rewriting `` `VdotCalculator` renamed to `PaceZoneIndexCalculator` `` as "legacy index calculator renamed to `PaceZoneIndexCalculator`" loses the specific identifier traceability — future readers doing `git log -S VdotCalculator` would hit commits this ROADMAP entry describes.

The project's own `backend/CLAUDE.md` uses the same pattern (referring to `FitnessEstimate.EstimatedVdot` and "the VDOT term" in its trademark disclaimer section), so ROADMAP is internally consistent.

backend/src/RunCoach.Api/Modules/Coaching/ContextAssembler.cs
```diff
@@ -520,26 +522,29 @@ private PromptSection BuildFitnessEstimateSection(FitnessEstimate fitness)
...
         {
-            sb.AppendLine(CultureInfo.InvariantCulture, $"Interval pace: {FormatTimeSpan(paces.IntervalPace.Value)} /km");
+            sb.AppendLine(CultureInfo.InvariantCulture, $"Interval pace: {FormatPace(paces.IntervalPace.Value)} /km");
         }
 
         if (paces.RepetitionPace.HasValue)
         {
-            sb.AppendLine(CultureInfo.InvariantCulture, $"Repetition pace: {FormatTimeSpan(paces.RepetitionPace.Value)} /km");
+            sb.AppendLine(CultureInfo.InvariantCulture, $"Repetition pace: {FormatPace(paces.RepetitionPace.Value)} /km");
         }
```
leehopper, line 549: **💡 [LOW] 🐛 FastRepetitionPace computed but never rendered into the LLM prompt**

`PaceZoneCalculator.CalculatePaces` always populates `FastRepetitionPace` (`PaceZoneCalculator.cs:71`), but this method has conditional blocks only for `EasyPaceRange`, `MarathonPace`, `ThresholdPace`, `IntervalPace`, and `RepetitionPace` — there is no block for `FastRepetitionPace`. The F-zone pace is computed on every call and then silently dropped; the LLM coach never sees it.

Also update `context-injection-v1.yaml` `training_paces.format` to list the new row, so the layout spec stays in sync.

```suggestion
        if (paces.RepetitionPace.HasValue)
        {
            sb.AppendLine(CultureInfo.InvariantCulture, $"Repetition pace: {FormatPace(paces.RepetitionPace.Value)} /km");
        }

        if (paces.FastRepetitionPace.HasValue)
        {
            sb.AppendLine(CultureInfo.InvariantCulture, $"Fast-repetition pace: {FormatPace(paces.FastRepetitionPace.Value)} /km");
        }
```

leehopper (commented):
Deep Review — 14 findings total (2 main, 12 suggestions). Posting the 4 highest-impact ones as line-anchored comments; the full report lives in the session.

🤖 Generated by [deep-review](https://github.com/anthropics/claude-code) · Reviewed up to: c3e538fe

backend/src/RunCoach.Api/Modules/Training/Models/PaceRange.cs
```diff
@@ -1,22 +1,28 @@
...
+        Slow = slow;
     }
 
-    public TimeSpan MinPerKm { get; init; }
+    /// <summary>Gets the faster (lower sec/km) end of the range.</summary>
+    public Pace Fast { get; init; }
 
-    public TimeSpan MaxPerKm { get; init; }
+    /// <summary>Gets the slower (higher sec/km) end of the range.</summary>
+    public Pace Slow { get; init; }
```
leehopper: **🟡 [MEDIUM] 🛡 `{ get; init; }` lets `with`-expressions silently bypass the Fast ≤ Slow invariant**

`PaceRange` validates `Fast.IsSlowerThan(Slow)` in its user-defined constructor, but both properties are declared `{ get; init; }`. C# `with` expressions invoke the compiler-synthesized copy constructor and then run `init` setters directly — skipping the user constructor entirely. So:

```csharp
var valid = new PaceRange(Pace.FromSecondsPerKm(300), Pace.FromSecondsPerKm(360));
var broken = valid with { Slow = Pace.FromSecondsPerKm(280) };
// broken.Fast = 300 s/km, broken.Slow = 280 s/km — invariant silently violated, no exception
```

`IntRange.cs:19-21` correctly uses `{ get; }` (no init), which blocks `with` mutations of those properties at compile time. Adopting the same pattern here removes the bypass.

No test currently exercises or documents this bypass (deep-review found no `with` tests against `PaceRange`), which is the companion finding `test-4`.

```suggestion
    /// <summary>Gets the faster (lower sec/km) end of the range.</summary>
    public Pace Fast { get; }

    /// <summary>Gets the slower (higher sec/km) end of the range.</summary>
    public Pace Slow { get; }
```

backend/tests/RunCoach.Api.Tests/SmokeTests.cs
```diff
@@ -5,9 +5,24 @@ namespace RunCoach.Api.Tests;
 
 public class SmokeTests(WebApplicationFactory<Program> factory) : IClassFixture<WebApplicationFactory<Program>>
 {
+    // Integration smoke test: boots the full API via WebApplicationFactory with Testcontainers
+    // Postgres. Opt-in only — set RUNCOACH_INTEGRATION_TESTS=1 to include (CI does this).
```
leehopper: **💡 [LOW] 📝 Comment says CI sets `RUNCOACH_INTEGRATION_TESTS=1` but no workflow does**

Grep across `.github/workflows/*.yml` returns zero matches for `RUNCOACH_INTEGRATION_TESTS`. The `ci.yml` job never sets the env var, so `IntegrationTestsEnabled` is always `false` in CI and `GetHealth_ReturnsOk` is permanently skipped.

Two reasonable fixes — pick whichever matches intent:

1. **Wire it up.** CI already has a `postgres` service container; adding `env: RUNCOACH_INTEGRATION_TESTS: "1"` on the `dotnet test` step completes the originally intended flow.
2. **Fix the comment.** If the smoke tests are deliberately opt-in for ad-hoc local runs and CI doesn't need them, update the parenthetical to say `(not run in CI)` and drop the misleading `(CI does this)` claim.

Leaving both as-is means the test appears to be part of CI coverage but isn't.

leehopper:
Deep-review follow-up comment that couldn't be anchored inline because the target line (`ContextAssemblerTests.cs:149`) isn't part of this PR's diff.

## 🐛 Fitness-section assertion passes coincidentally, not semantically

**File:** `backend/tests/RunCoach.Api.Tests/Modules/Coaching/ContextAssemblerTests.cs:149`

Production code in `ContextAssembler.BuildFitnessEstimateSection` now emits `"Pace-zone index: {value}"` as the label (no `VDOT`). The assertion `fitnessSection.Content.Should().Contain("VDOT")` passes only because `TestProfiles.Lee()` stores `AssessmentBasis = "10K race time of 48:00 (2026-02-15) -> VDOT {vdot}"` — so the literal `VDOT` appears in the rendered section as *user data*, not as a label. The assertion looks like a trademark-regression guard but no longer guards anything.

If/when the `AssessmentBasis` string is scrubbed (tracked as a followup), this test will fail for the wrong reason.

**Fix:**

1. Change the assertion to `Contain("Pace-zone index:")` so it validates the actual label.
2. While you're here, scrub `"VDOT"` from `TestProfiles.Lee().AssessmentBasis` — it is a third surviving VDOT identifier not yet in ROADMAP's follow-up list.

```csharp
fitnessSection.Content.Should().Contain("Pace-zone index:");
```

🤖 Generated by deep-review · Reviewed up to: c3e538fe

backend/src/RunCoach.Api/Modules/Coaching/ContextAssembler.cs
leehopper: Fixed in 0f439a6: added the missing `if (paces.FastRepetitionPace.HasValue)` block in `BuildTrainingPacesSection` (ContextAssembler.cs:549-552) and extended the `training_paces.format` row in `context-injection-v1.yaml`. A new test `AssembleAsync_ProfileWithFastRepetitionPace_RendersFastRepetitionPace` pins the render path, and the existing beginner negative-assertion test now also confirms absence. No eval-cache regeneration needed — `TestProfiles` use the lookup-table bridge which leaves `FastRepetitionPace` null.

backend/src/RunCoach.Api/Modules/Training/Models/PaceRange.cs
leehopper: Fixed in 860d837: dropped `init` on both `Fast` and `Slow` so they are now `{ get; }` only, matching the `IntRange` pattern. `with`-expression reassignment of either property is now a compile-time error. The full test suite still passes (550 tests), confirming no existing consumer was using `with` to mutate these properties. A regression test `WithExpression_CannotReassignFastOrSlow` documents the guarantee for future readers.

backend/tests/RunCoach.Api.Tests/SmokeTests.cs
leehopper: Fixed in d990e3f: updated the comment to say "Not currently wired in CI (the workflow doesn't set the env var)" so it matches reality. Chose the comment-fix path rather than wiring the env var — enabling integration runs in CI is a separate decision that deserves its own PR (the postgres service is already wired but the container image and health-check timing haven't been validated end-to-end yet).

leehopper:
**Fixed in 936196c:** changed the assertion from `Contain("VDOT")` to `Contain("Pace-zone index:")` so it validates the actual production label instead of relying on a coincidental string match against user-data in `AssessmentBasis`.

Scrubbing the remaining `"VDOT"` literal from `TestProfiles.Lee().AssessmentBasis` is deferred — that text is part of the fixture prompt content and changing it would invalidate committed eval-cache fixtures, requiring a re-record.

PR #50, opened 2026-04-22
feat(slice-0): t02 auth api — register / login / logout / me / xsrf

leehopper:
## Summary

Slice 0 Unit 2 — the auth substrate every later slice's authorization policies depend on. Adds a hand-rolled `AuthController` with `/xsrf`, `/register`, `/login`, `/me`, `/logout`; Identity Core + JWT-bearer opt-in schemes wired via a `CookieOrBearer` authorization policy; antiforgery double-cookie pattern via the `UseAntiforgery()` middleware; RFC 7807 ProblemDetails error pipeline; and a 13-case integration test matrix against the shared Testcontainers Postgres (T01.5 `AssemblyFixture`). Closes out T02.1 → T02.5 on the task board.

- **Browser session is the `__Host-RunCoach` Identity application cookie** (HttpOnly, Secure, SameSite=Lax, 14-day sliding) — no JWT issued to the browser per DEC-044.
- **JWT bearer is registered but dormant** per DEC-051 so the future iOS shim lands additively against the same `CookieOrBearer` policy.
- **Antiforgery uses `[RequireAntiforgeryToken]` + `AddControllers()` + a 3-line bridge middleware** per DEC-055 (R-062 research) — middleware reads `IAntiforgeryMetadata`, bridge converts `IAntiforgeryValidationFeature.IsValid == false` to a clean 400 on the MVC pipeline. `[ValidateAntiForgeryToken]` + `AddControllersWithViews()` was ruled out (drags in an unused Razor view engine).
- **Timing-safe login** per DEC-053 (R-059): `UserManager.FindByEmailAsync` + cached dummy-hash `VerifyHashedPassword` on the unknown-user branch so the wrong-password and unknown-user paths are indistinguishable in wall-clock.
- **`IdentityResult.Errors` → DTO-property-keyed `ValidationProblemDetails`** per DEC-052 (R-058), with duplicate-email short-circuited to a generic 409 to preserve enumeration resistance.
- **`/xsrf` writes the SPA-readable `__Host-Xsrf-Request` cookie** alongside the framework `__Host-Xsrf` cookie per DEC-054. RTK Query will read it in T03.1.

## Decisions captured

- **DEC-050** — `__Host-RunCoach` cookie posture (HttpOnly, Secure, SameSite=Lax, 14-day).
- **DEC-051** — `AddJwtBearer` registered as a non-default opt-in scheme; `JwtAuthOptionsValidator` is gated on `!IsDevelopment()`.
- **DEC-052** — Identity error mapping to DTO-property-keyed ValidationProblemDetails.
- **DEC-053** — Timing-safe login with manual dummy-hash verification.
- **DEC-054** — `/xsrf` SPA-readable cookie shape + `[ApiController]` automatic 401/403.
- **DEC-055** — `[RequireAntiforgeryToken]` + `AddControllers()` + bridge middleware (supersedes DEC-054 bullet 3).

## Research artifacts integrated

- R-056 `batch-19a-httpsredirection-webapplicationfactory.md` — `UseHttpsRedirection` gate + `__Host-` cookie composition.
- R-057 `batch-19b-jwtbearer-opt-in-registration.md` — dormant JWT scheme posture.
- R-058 `batch-19c-identity-result-to-validationproblemdetails.md` — error-shape translation.
- R-059 `batch-19d-signinmanager-timing-safety.md` — unknown-user timing parity.
- R-062 `batch-19e-antiforgery-attribute-net10.md` — attribute selection + middleware bridge.

## Test plan

- [x] Unit suite green — `dotnet test tests/RunCoach.Api.Tests` reports **601 / 601** pass.
- [x] T02.5 integration matrix covers every spec variant (13 cases): `/xsrf` 204 + cookie, `/register` happy-path / duplicate / weak-password / missing-antiforgery, `/login` happy-path / wrong-password / unknown-user (byte-identical ProblemDetails contract), `/me` authenticated with DB-mutation proof / anonymous, `/logout` authenticated / anonymous / missing-antiforgery.
- [x] Startup smoke tests still green (DI graph, DataProtection round-trip, `/health`, etc.).
- [ ] `/review-pr` locally before merge.
- [ ] Manual smoke through `dotnet run` + curl for `/xsrf` → `/register` → `/login` → `/me` → `/logout` once PR is ready to merge.

## Follow-ups found

None blocking merge. All auth-module research prompts resolved; every DEC claim verified against primary sources or direct source-code inspection.

- [ ] T03.x frontend consumes this contract. T03.1 wires RTK Query's `baseQuery` to read `__Host-Xsrf-Request` and echo it as `X-XSRF-TOKEN`.
- [ ] Consider promoting the antiforgery bridge middleware (currently inline in `Program.cs`) to a named class (`AntiforgeryValidationBridgeMiddleware`) if a future change adds ProblemDetails body emission or audit logging. DEC-055 documents the growth trigger.
- [ ] `[RequireAntiforgeryToken]` is not yet documented on Microsoft Learn for MVC controllers (`dotnet/AspNetCore.Docs#33740` tracks the gap). Re-verify the pattern when that doc lands in case Microsoft's recommended shape changes.

## Release Notes

* **New Features**
  * Added user registration with email and password validation
  * Added login endpoint with secure session-based authentication
  * Added profile endpoint to retrieve authenticated user details
  * Added antiforgery protection for API mutations (registration, login, logout)
  * Added optional JWT bearer token authentication for future use

* **Documentation**
  * Added comprehensive contributing guide with local development setup instructions
  * Updated README to clarify HTTPS setup requirements


backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
```diff
@@ -0,0 +1,189 @@
...
+        Response.Cookies.Append(
+            RequestTokenCookieName,
+            tokens.RequestToken!,
+            new CookieOptions
+            {
+                HttpOnly = false,
+                Secure = true,
+                SameSite = SameSiteMode.Lax,
+                Path = "/",
+            });
```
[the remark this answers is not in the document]
leehopper: **Intentional per DEC-054 — pushback.**

This cookie is the SPA-readable half of the antiforgery double-submit pair:

- `__Host-Xsrf` (framework-managed): `HttpOnly=true` — stays in the browser jar.
- `__Host-Xsrf-Request` (this write): `HttpOnly=false` — the SPA reads it and echoes the value in the `X-XSRF-TOKEN` request header. That readability is the whole point of the cookie; setting `HttpOnly=true` would break the SPA's ability to form unsafe requests.

See DEC-054 and Microsoft Learn's SPA antiforgery pattern. Both cookies carry `__Host-` prefix + `Secure` + `Path=/` per RFC 6265bis. The CSRF surface is closed by the framework cookie (server-bound, HttpOnly) — the SPA-readable companion is the token-echo channel.

Will address the scanner signal with an inline `[SuppressMessage]` + justification comment in the next commit, and mark the Sonar hotspot "Safe — reviewed" on the dashboard.

backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
```diff
@@ -0,0 +1,203 @@
...
```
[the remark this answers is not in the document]
leehopper: Same root cause / same disposition as [#discussion_r3127398017](https://github.com/leehopper/personal-running-coach/pull/50#discussion_r3127398017) — this is the SPA-readable half of the antiforgery double-submit pair per DEC-054 and must be JS-readable. Will address the scanner signal with an inline `[SuppressMessage]` + justification in the next commit.

backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        var response = new AuthResponse(user.Id, user.Email!);
+        return StatusCode(StatusCodes.Status201Created, response);
+    }
+
+    [HttpPost("login")]
+    [AllowAnonymous]
+    [ProducesResponseType(typeof(AuthResponse), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status401Unauthorized)]
+    public async Task<IActionResult> Login([FromBody] LoginRequest request)
```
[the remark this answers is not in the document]
leehopper: Addressed in 388c3cb — added `[RequireAntiforgeryToken]` to the `Login` action (matches Register / Logout). New test `Login_MissingAntiforgeryToken_Returns_400` pins the behavior; Cases 6 / 7 / 8 + the `LoginAsync` helper were updated to prime antiforgery before POSTing.

backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
```diff
@@ -0,0 +1,189 @@
...
+    public async Task<IActionResult> Me()
+    {
+        var userId = User.FindFirstValue(ClaimTypes.NameIdentifier);
+        if (string.IsNullOrEmpty(userId))
+        {
+            return Unauthorized();
+        }
+
+        var user = await userManager.FindByIdAsync(userId);
+        if (user is null)
```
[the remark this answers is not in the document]
leehopper: Addressed in 388c3cb — `Me()` now falls back from `ClaimTypes.NameIdentifier` to `JwtRegisteredClaimNames.Sub` when the former is null, with an inline comment explaining why (`MapInboundClaims=false` in Program.cs keeps JWT claim names raw). New integration test `Me_WithBearerSubClaim_Returns_200_WithExpectedEmail` mints an HS256 bearer carrying only `sub`, hits `/me`, asserts the fallback resolves to the registered user. The test fixture now seeds deterministic `Auth:Jwt` env vars so bearer-path assertions don't need a separate WebApplicationFactory subclass.

backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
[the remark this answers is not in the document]
leehopper: Scanner suppression applied in 388c3cb — `IssueAntiforgeryTokens` now carries `[SuppressMessage]` entries for `cs/web/cookie-httponly-not-set` (CodeQL) and `S2092` (Sonar) with justifications pointing at DEC-054. The `HttpOnly=false` line remains as designed (SPA-readable double-submit companion).

backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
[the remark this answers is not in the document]
leehopper: Scanner suppression applied in 388c3cb — see [#discussion_r3127398017](https://github.com/leehopper/personal-running-coach/pull/50#discussion_r3127398017) reply for details.

backend/src/RunCoach.Api/Modules/Identity/IdentityResultExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        foreach (var error in result.Errors)
+        {
+            var key = IdentityErrorCodeMapper.Map(error).PropertyName switch
+            {
+                IdentityErrorBuckets.Password => nameof(RegisterRequest.Password).ToCamelCase(),
+                IdentityErrorBuckets.Email => nameof(RegisterRequest.Email).ToCamelCase(),
+                IdentityErrorBuckets.UserName => nameof(RegisterRequest.Email).ToCamelCase(),
+                _ => string.Empty,
+            };
+            controller.ModelState.AddModelError(key, error.Description);
```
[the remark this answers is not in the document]
leehopper: Addressed in 549ce99 — the default switch arm in `IdentityResultExtensions.ToRegistrationActionResult` now routes through `IdentityErrorBuckets.General` instead of returning `string.Empty`. Non-field Identity errors (`DefaultError`, `ConcurrencyFailure`, and any unknown code the mapper sees in the future) surface under the canonical `general` bucket. Coverage on both files is also addressed: new `IdentityErrorCodeMapperTests` (23 explicit mapping assertions) + `IdentityResultExtensionsTests` (11 unit tests covering every branch, including the general-bucket path for `DefaultError` and unknown codes).

docs/research/artifacts/batch-19a-httpsredirection-webapplicationfactory.md
```diff
@@ -0,0 +1,275 @@
+# UseHttpsRedirection + `__Host-` cookies + WebApplicationFactory: the canonical 2026 pattern
+
+**Research artifact** — drop into `docs/research/artifacts/` as `batch-15b-https-redirection-host-cookie-waf.md`. Dates verified against primary sources as of April 21, 2026; .NET 10 GA shipped November 2025.
```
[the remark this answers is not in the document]
leehopper: Addressed in 9198b96 — filename self-reference now points at `batch-19a-httpsredirection-webapplicationfactory.md` (the actual stored filename).

docs/research/artifacts/batch-19b-jwtbearer-opt-in-registration.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Test-host posture (today, cookie-only)
+
+```csharp
+public sealed class RunCoachWebApplicationFactory : WebApplicationFactory<Program>
+{
+    protected override void ConfigureWebHost(IWebHostBuilder builder)
+    {
+        builder.UseEnvironment("Testing");  // skips ValidateOnStart branch
+        // No auth overrides. Cookie path works. JWT registered but rejects
```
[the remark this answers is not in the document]
leehopper: Addressed in 9198b96 — snippet changed to `UseEnvironment("Development")` with an inline comment explaining that `IsDevelopment()` is only true for "Development", so `"Testing"` would still fire `ValidateOnStart` in CI where `Auth:Jwt` is unset.

docs/research/prompts/batch-19d-signinmanager-timing-safety.md
```diff
@@ -0,0 +1,73 @@
...
+
+The spec's fallback pattern is:
+
+```csharp
+var user = await userManager.FindByEmailAsync(email);
+if (user is null)
+{
+    // dummy hash to match the timing of a real password verification
+    await userManager.PasswordHasher.HashPasswordAsync(default, password);
+    return Unauthorized(GenericProblemDetails());
```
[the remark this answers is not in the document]
leehopper: Addressed in 9198b96 — snippet switched to the real synchronous `IPasswordHasher<TUser>.HashPassword` API (no `HashPasswordAsync` exists) with a discard to pin the timing-mitigation shape.

docs/research/prompts/batch-19e-antiforgery-attribute-net10.md
```diff
@@ -0,0 +1,65 @@
...
+
+**Sub-questions:**
+
+1. **Attribute selection.** For a pure JSON API hosted in MVC controllers (no Razor views, no Razor Pages) using the `UseAntiforgery` middleware — which attribute does Microsoft's current guidance (Learn / dotnet/aspnetcore samples / ASP.NET Core team posts) recommend? What primary source (not a blog) supports the recommendation?
+2. **`[RequireAntiforgeryToken]` on MVC controllers.** `Microsoft.AspNetCore.Http.RequireAntiforgeryTokenAttribute` is documented for Minimal APIs. Does placing it on an MVC controller action emit the same `IAntiforgeryMetadata` such that `UseAntiforgery` validates the request, without requiring any MVC filter registration? Any caveats around filter ordering, `[AllowAnonymous]`, or `[IgnoreAntiforgeryToken]` carve-outs?
+3. **`[IgnoreAntiforgeryToken]` carve-out.** If we switch to `[RequireAntiforgeryToken]` globally, how do we exempt the `GET /xsrf` endpoint from validation? Does `[IgnoreAntiforgeryToken]` (MVC) produce the corresponding `IAntiforgeryMetadata(required: false)`, or is a different mechanism required (e.g., `DisableAntiforgery()` endpoint convention)?
```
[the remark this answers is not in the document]
leehopper: Addressed in 9198b96 — namespace corrected from `Microsoft.AspNetCore.Http` to `Microsoft.AspNetCore.Antiforgery` (where `RequireAntiforgeryTokenAttribute` actually lives per .NET 10).

docs/research/research-queue.md
```diff
@@ -94,6 +94,13 @@ Research is batched by dependency. Later batches depend on findings from earlier
...
...AspNetCore.Mvc.ValidateAntiForgeryTokenAttribute` (requires `AddViews()`-backed filter via `AddControllersWithViews()`) or `Microsoft.AspNetCore.Antiforgery.RequireAntiforgeryTokenAttribute` (endpoint-metadata path, no MVC filter registration)? Confirm / reject the parent task's claim that `[ValidateAntiForgeryToken]` is "broken with `UseAntiforgery` in .NET 10", diagnose whether dual-validation occurs when both filter + middleware are in the pipeline, resolve the `[IgnoreAntiforgeryToken]` vs `DisableAntiforgery()` carve-out for `GET /xsrf`, and pin the `AddControllers*` registration shape. | **T02 PR-packaging follow-up.** Integrated 2026-04-22 as DEC-055. Adopted `[RequireAntiforgeryToken]` + `AddControllers()`; added a one-line middleware after `UseAntiforgery()` that converts `IAntiforgeryValidationFeature.IsValid == false` → 400 (the research's "short-circuits with 400" claim was verified against aspnetcore `release/10.0` source and found to be wrong for MVC — the middleware sets the feature and calls `_next`; Minimal APIs bridge this to 400 via form-binding, MVC does not). All 601 / 601 tests green on the new pattern. | Integrated | batch-19e-antiforgery-attribute-net10.md |
```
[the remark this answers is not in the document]
leehopper: Addressed in 9198b96 — R-062 status cell now `Done`. The integration context (DEC-055, 601 / 601 test result) was merged into the question cell so the queue keeps the `Queued`/`Done` vocabulary clean.

ROADMAP.md
```diff
@@ -1,9 +1,9 @@
...
+**Active slice:** Slice 0 (Foundation) — Unit 1 persistence substrate shipped (PR #49 merged 2026-04-21, commit `46348fc`; all 14 CI checks green on merge). DEC-048 (Marten `IntegrateWithWolverine` as sole envelope-storage wiring) and DEC-049 (disable host-config reload on macOS arm64 + drop manual `MapWolverineEnvelopeStorage`) both landed with the merge; full `WebApplicationFactory<Program>` fixture + six SUT-host smoke tests carry 582 passing / 0 failing / 0 skipped locally.
+**Next step:** Begin Unit 2 (T02.x — Auth API) starting with T02.1 (JwtBearer package + Identity Core registration + password policy + `CookieOrBearer` name), then T02.2 → T02.5 (auth schemes + antiforgery + middleware, global exception handler, `AuthController`, integration-test matrix). Slice 0 acceptance remains gated on Unit 2 (auth endpoints) + Unit 3 (frontend auth UX). DEC-044 (cookie-not-JWT browser auth), DEC-045 (Aspire deferred to MVP-1), DEC-046 (SOPS + Postgres-DataProtection secrets), and DEC-047 (onboarding event-source pattern) remain the architectural anchors for subsequent slices.
```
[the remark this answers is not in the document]
leehopper: Addressed in 9198b96 — `ROADMAP.md` + `docs/plans/mvp-0-cycle/cycle-plan.md` Status blocks updated to reflect Unit 2 T02.1–T02.6 in-flight on PR #50 (register / login / me / logout / xsrf + DEC-050 through DEC-056 + 638 tests green). Next step now points at Unit 3 frontend auth UX.

docs/research/artifacts/batch-20a-dotnet10-sdk-container-sigill-apple-silicon.md
```diff
@@ -0,0 +1,313 @@
...
++FROM mcr.microsoft.com/dotnet/sdk:10.0@sha256:8a90a473da5205a16979de99d2fc20975e922c68304f5c79d564e666dc3982fc AS build
++
++# Workaround for dotnet/runtime#122608 (milestoned .NET 11).
++# Apple M3/M4/M5 under macOS Virtualization.framework expose HWCAP2_SVE2 but only
++# implement SVE2 inside SME streaming mode; RyuJIT emits non-streaming SVE2 => SIGILL.
++# Disabling these two CoreCLR knobs is a no-op on x86_64 (ignored) and on arm64
++# hardware that does not falsely advertise SVE2, so this is safe for GitHub Actions
++# ubuntu-latest runners and for production Linux arm64 hosts (Graviton, Ampere).
++ENV DOTNET_EnableHWIntrinsic_Arm64Sve=0 \
++    DOTNET_EnableHWIntrinsic_Arm64Sve2=0
```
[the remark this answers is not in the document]
leehopper: Addressed in 9198b96 — env-var names corrected to the `jitconfigvalues.h` canonical spelling (`DOTNET_EnableArm64Sve*` vs the previous `DOTNET_EnableHWIntrinsic_Arm64Sve*`). Also added a preamble integration note pointing at R-064's empirical verification that these knobs do not cure the fault — names were correct but the crash sits in a pre-CLRConfig startup probe. DEC-056 posture unchanged.

backend/src/RunCoach.Api/Modules/Identity/Contracts/LoginRequestDto.cs
```diff
@@ -0,0 +1,17 @@
...
+
+/// <summary>
+/// Request body for <c>POST /api/v1/auth/login</c>. Validation here is
+/// deliberately permissive — format mismatches must produce the same 401
+/// <see cref="Microsoft.AspNetCore.Mvc.ProblemDetails"/> as a genuinely
+/// unknown user, not a 400, so an enumerator cannot distinguish the two.
+/// Only <c>[Required]</c> is applied to reject missing fields via model
+/// binding before the timing-mitigation path runs.
+/// </summary>
+public sealed record LoginRequestDto(
```
leehopper, line 13: Addressed in f24a386 (tracking thread for the review-body nitpick "rename `LoginRequest` → `LoginRequestDto`"). File + record + all references renamed to match the `Dto` suffix convention from backend/CLAUDE.md.

backend/src/RunCoach.Api/Modules/Identity/Contracts/RegisterRequestDto.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace RunCoach.Api.Modules.Identity.Contracts;
+
+/// <summary>
+/// Request body for <c>POST /api/v1/auth/register</c>. DataAnnotations enforce
+/// format + basic length so the common "too short" / malformed-email path
+/// short-circuits through <c>[ApiController]</c> auto-400 before reaching
+/// Identity (DEC-052). Identity remains source-of-truth for character-class
+/// rules and uniqueness.
+/// </summary>
+public sealed record RegisterRequestDto(
```
leehopper, line 12: Addressed in f24a386 (tracking thread for the review-body nitpick "rename `RegisterRequest` → `RegisterRequestDto`"). File + record + refs renamed. Callers + tests + `IdentityResultExtensions` all updated.

backend/src/RunCoach.Api/Modules/Identity/Contracts/AuthResponseDto.cs
```diff
@@ -0,0 +1,8 @@
+namespace RunCoach.Api.Modules.Identity.Contracts;
+
+/// <summary>
+/// Response body for <c>POST /api/v1/auth/register</c>, <c>POST /api/v1/auth/login</c>,
+/// and <c>GET /api/v1/auth/me</c>. No token field — the session IS the
+/// <c>__Host-RunCoach</c> application cookie set by <c>SignInManager</c>.
+/// </summary>
+public sealed record AuthResponseDto(Guid UserId, string Email);
```
leehopper, line 8: Addressed in f24a386 (tracking thread for the review-body nitpick "rename `AuthResponse` → `AuthResponseDto`"). File + record + refs renamed. `AuthController` now returns `AuthResponseDto` on register / login / me.

backend/src/RunCoach.Api/Infrastructure/JwtAuthOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+/// </summary>
+/// <remarks>
+/// Slice 0 has no endpoint that accepts the bearer scheme. Absent values are
+/// tolerated in Development and CI because <c>ValidateOnStart</c> is only
+/// called outside <c>IsDevelopment()</c>; the JWT handler's strict-always
+/// validation (R-057) then rejects every incoming token with
+/// <c>IDX10500</c> until real config lands. In Production / Staging,
+/// startup fails fast if any required value is missing.
+/// </remarks>
+public sealed record JwtAuthOptions
```
leehopper, line 19: Addressed in f24a386 (tracking thread for the review-body nitpick "`JwtAuthOptions` should be a record with non-null string defaults"). `sealed class` → `sealed record`. `Issuer`/`Audience`/`SigningKey`/`KeyId` are now `string` with `= string.Empty` init defaults per backend/CLAUDE.md's "properties initialized with default non-null values" rule. `[Required]` + `[MinLength]` annotations preserved; Program.cs's `o.KeyId ?? "current"` collapsed into an `IsNullOrEmpty` check now that the property is non-null.

backend/src/RunCoach.Api/Modules/Identity/IdentityErrorMapping.cs
```diff
@@ -0,0 +1,9 @@
+namespace RunCoach.Api.Modules.Identity;
+
+/// <summary>
+/// The output of <see cref="IdentityErrorCodeMapper.Map"/> — a pair of the
+/// DTO-property bucket name the error should surface under and the HTTP
+/// semantic class the error represents. Kept as a dedicated type in its
+/// own file per the project one-type-per-file rule.
```
leehopper: Addressed in f24a386 (tracking thread for the review-body nitpick "extract `IdentityErrorCodeMapper.Mapping` to its own file"). Public `readonly record struct` moved out of the mapper file into `IdentityErrorMapping.cs` + renamed to `IdentityErrorMapping`. The one-type-per-file rule's "internal serialization helper" exception didn't apply — the type is public. Mapper's `Map` return type updated.

backend/src/RunCoach.Api/Program.cs
```diff
@@ -167,13 +346,114 @@
...
+// IAntiforgeryValidationFeature but does NOT short-circuit — Minimal API
+// form-binding converts that to a 400 automatically; MVC controllers do
+// not. Without this bridge, a state-changing POST with a missing or
+// stale token either 500s (form-binding path hits
+// HandleUncheckedAntiforgeryValidationFeature) or silently succeeds
+// (no-body POST has nothing to trigger the check). See
+// dotnet/AspNetCore.Docs#33740. Per-endpoint `[RequireAntiforgeryToken]`
+// metadata is the canonical attribute; this middleware is the matching
+// gate for the MVC pipeline.
+app.Use(async (ctx, next) =>
```
leehopper: Addressed in 9a12a09 (tracking thread for the review-body nitpick "antiforgery bridge should return ProblemDetails instead of empty 400"). Bridge middleware now resolves `IProblemDetailsService` and calls `TryWriteAsync` with a full RFC 7807 body — type `https://runcoach.app/problems/antiforgery-validation-failed`, title, and a detail that points callers at `GET /api/v1/auth/xsrf` for re-seeding. Contract parity with the rest of the 4xx auth stack. Test `Register_MissingAntiforgeryToken_Returns_400_ProblemDetails` pins the new shape.

backend/tests/RunCoach.Api.Tests/Modules/Identity/AuthControllerIntegrationTests.cs
```diff
@@ -0,0 +1,625 @@
...
+/// <summary>
+/// Integration matrix for the five Slice 0 auth endpoints
+/// (<c>/xsrf</c>, <c>/register</c>, <c>/login</c>, <c>/me</c>, <c>/logout</c>)
+/// against the shared Testcontainers Postgres. Every case drives the real
+/// <see cref="RunCoachAppFactory"/> SUT — no mocks — and uses
+/// <see cref="CookieContainerHandler"/> so the antiforgery + session cookie
+/// lifecycle is observable end-to-end.
+/// </summary>
+[Trait("Category", "Integration")]
+public class AuthControllerIntegrationTests(RunCoachAppFactory factory) : DbBackedIntegrationTestBase(factory)
```
leehopper, line 30: Addressed in f24a386 (tracking thread for the review-body nitpick "rename `AuthControllerTests` → `AuthControllerIntegrationTests`"). File + class renamed via `git mv` to preserve history. Suite inherits `DbBackedIntegrationTestBase` + carries `[Trait("Category", "Integration")]` — naming now matches backend/CLAUDE.md's `{ClassName}IntegrationTests.cs` convention.

backend/tests/RunCoach.Api.Tests/Modules/Identity/AuthControllerIntegrationTests.cs
```diff
@@ -0,0 +1,625 @@
...
+            .And.Be(HttpStatusCode.Unauthorized);
+
+        // … and byte-equivalent JSON bodies after stripping the per-request
+        // `traceId` extension. Field-by-field comparison on the typed
+        // ProblemDetails surface would let any future ProblemDetails
+        // extension (a role flag, a retry-after hint) diverge between
+        // branches unnoticed — deep equality on the normalized JSON catches
+        // every property the serializer produces, not just the four the
+        // type surface knows about.
+        var expectedBody = await ReadNormalizedProblemJsonAsync(wrongPasswordResponse);
```
leehopper, line 319: Addressed in 810152a (tracking thread for the review-body nitpick "compare the normalized 401 bodies end-to-end, not just Type/Title/Status/Detail"). `ReadNormalizedProblemJsonAsync` helper parses the response body as `JsonNode`, strips the per-request `traceId` extension, and returns the serialized remainder. The two bodies are then compared byte-for-byte via `Should().Be(...)`, which catches every property the serializer produces — not just the four the typed `ProblemDetails` surface knows about. A parallel assertion pulls `type` back out to pin the `InvalidCredentials` helper-drives-both-branches guarantee.

backend/src/RunCoach.Api/Program.cs
```diff
@@ -93,6 +101,161 @@
...
+// `AddIdentityCookies` returns an `IdentityCookiesBuilder`, so the two calls
+// cannot be chained. `TokenValidationParameters` are composed by the options
+// pipeline below, not here, so env-var overrides applied at test-init time
+// are picked up.
+var authBuilder = builder.Services.AddAuthentication(IdentityConstants.ApplicationScheme);
+authBuilder.AddIdentityCookies(options =>
+{
+    options.ApplicationCookie?.Configure(cookie =>
+    {
+        cookie.Cookie.Name = AuthCookieNames.Session;
```
leehopper, line 159: Addressed in 9a12a09 (tracking thread for the review-body nitpick "centralize the auth cookie / header names"). New `Infrastructure/AuthCookieNames.cs` is now the single source of truth for `__Host-RunCoach` / `__Host-Xsrf` / `__Host-Xsrf-Request` / `X-XSRF-TOKEN`. All five call sites in Program.cs + AuthController + the Swagger interceptor + the integration test constants reference the static fields instead of string literals, so a rename is a one-commit change rather than five-file drift.

backend/src/RunCoach.Api/Program.cs
```diff
@@ -136,11 +299,27 @@
...
+builder.Services.AddExceptionHandler<ErrorHandlingMiddleware>();
+
+// CORS — explicit origin is mandatory because `AllowCredentials()` is
+// incompatible with `AllowAnyOrigin()` (per the CORS spec, and ASP.NET Core
+// throws at the first preflight). HTTPS-only because the `__Host-RunCoach`
+// application cookie is `Secure`-flagged and only transmitted over TLS.
+// `AllowCredentials` is required so the browser includes the cookie (and
+// echoes the `X-XSRF-TOKEN` header) on cross-origin API calls from the Vite
+// dev server.
 builder.Services.AddCors(options =>
```
leehopper, line 278: **Deferred — pushback.**

Fair direction, but premature. The suggestion turns on "the first non-local frontend URL will fail browser auth until the API is rebuilt" — that assumes a hosted environment which does not yet exist. MVP-0 has zero production deploy targets (see cycle-plan § Captured During Cycle, "production-deployment-topology"). Wiring a `SpaCorsOptions` record now means picking a config contract shape (single-host list? per-environment split? reverse-proxy-aware?) against zero hosting facts, which is the exact YAGNI shape the cycle-plan's deferred-items section exists to prevent.

Will bundle this with the `ForwardedHeaders` middleware + secrets-bootstrap work at production-target commit time (pre-MVP-1). Captured on the cycle-plan's Captured-During-Cycle table so the follow-up doesn't get lost.

CONTRIBUTING.md
```diff
@@ -0,0 +1,254 @@
...
+posture — use at your own discretion:
+
+1. **Docker Desktop ≥ 4.39** — masks the offending CPU features at the
+   LinuxKit layer per [docker/for-mac#7583]. Install, `docker context use
+   desktop-linux`, then `tilt up` works. Trade-offs: Docker Business
+   subscription applies to orgs with >250 employees or >$10M revenue,
+   telemetry on by default, proprietary closed-source.
+2. **Colima with QEMU + pre-SVE CPU model** — removes SVE2/SME advertisement
+   before the guest kernel sees it, per [lima-vm#3032]:
+   ```bash
```
leehopper, line 163: Addressed in 9198b96 (tracking thread for the review-body nitpick "add language specifier to fenced code block"). Opening fence is now ````bash``` so markdownlint stops flagging MD040.

backend/Dockerfile
```diff
@@ -9,9 +48,16 @@ RUN dotnet restore RunCoach.slnx
...
+FROM mcr.microsoft.com/dotnet/aspnet:10.0@sha256:55e37c7795bfaf6b9cc5d77c155811d9569f529d86e20647704bc1d7dd9741d4 AS runtime
+
 WORKDIR /app
 COPY --from=build /app/publish .
+
+# Default runtime posture is plain HTTP on :5000 — a minimal baseline that
+# builds run in any CI or non-Compose context. For local Docker Compose dev
+# `docker-compose.yml` overrides these defaults to terminate HTTPS on :5001
+# using a dev cert mounted from the host, which is what `__Host-` cookies
+# and antiforgery require.
```
leehopper, line 60: Addressed in 9198b96 (tracking thread for the review-body nitpick "note that docker-compose.yml overrides these defaults"). Inline comment now explains that the plain-HTTP :5000 defaults are a minimal baseline for CI / non-Compose contexts, and that `docker-compose.yml` overrides to HTTPS :5001 with a dev cert for local browser-auth work.

docs/research/prompts/batch-20a-dotnet10-sdk-container-sigill-apple-silicon.md
```diff
@@ -0,0 +1,99 @@
...
+
+## PROMPT
+
+**Research Topic:** What is the canonical, durable 2026 setup for containerized `dotnet restore` of a realistic .NET 10 solution (Marten 8.x, Wolverine 5.x, Microsoft.Extensions.AI preview, Anthropic SDK, OpenTelemetry, Testcontainers) running on Apple Silicon under Colima — such that `tilt up` completes reliably, fast, and without SIGILL? Isolate the faulting code path, prescribe a fix that doesn't require every contributor to rediscover the workaround, and pin a known-good SDK image digest the repo can trust.
+
+## Context
+
+RunCoach is an ASP.NET Core 10 app with a solo-dev Docker Compose + Tilt loop. Tilt's `api` resource builds `backend/Dockerfile` which starts `FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build` and runs `dotnet restore RunCoach.slnx`. On a developer's macOS arm64 box the restore fails with:
+
+```text
```
leehopper, line 17: Addressed in 9198b96 (tracking thread for the review-body nitpick "add language specifier to error output code block"). Opening fence is now ````text```.

docs/research/artifacts/batch-20b-sve2-sigill-second-opinion-verification.md
```diff
@@ -0,0 +1,151 @@
...
...e Podman maintainer-level analysis of the identical bug class (containers/podman#28312, pgib, 2026-03-17) explicitly identifies **SME/SME2 features** — not SVE2 proper — as co-triggers on Apple M4/M5 under VZ-backed Linux: the guest `/proc/cpuinfo` advertises `sve2 sme sme2 sme2p1 smei16i32 smebi32i32 smef16f16` etc., and Apple Silicon implements SME/SME2 in streaming mode only but does not execute non-streaming SVE/SVE2. **CoreCLR 10.0 has no `DOTNET_EnableArm64Sme*` knob at all** — SME intrinsics are a .NET 11 work item per dotnet/runtime#121787. This means that even if R-063 had spelled the SVE knobs with perfect casing — and per `src/coreclr/jit/jitconfigvalues.h` the spelling `DOTNET_EnableArm64Sve` / `DOTNET_EnableArm64Sve2` was already correct and case-sensitive — there may be no env-var surface in 10.0.203 that disables the actual crashing code path, because the faulting instructions are plausibly in a pre-CLRConfig startup probe (as in the near-identical OpenJDK bug JDK-8345296) or are SME2 instructions with no user-facing gate. This explains empirically why three increasingly aggressive knobs (`Sve=0`, `Sve2=0`, `EnableHWIntrinsic=0`) all failed on the reference hardware.
```
leehopper, line 7: Addressed in 9198b96 (tracking thread for the review-body nitpick "capitalize 'Apple Silicon'"). All occurrences in this artifact use the proper-noun capitalization.

backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
```diff
@@ -0,0 +1,233 @@
...
+        // the claim-mapping surface.
+        var userId =
+            User.FindFirstValue(ClaimTypes.NameIdentifier)
+            ?? User.FindFirstValue(JwtRegisteredClaimNames.Sub);
+        if (string.IsNullOrEmpty(userId))
+        {
+            return Unauthorized();
+        }
+
+        var user = await userManager.FindByIdAsync(userId);
```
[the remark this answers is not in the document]
leehopper: Addressed in db6de3d — `Me()` now guards the resolved `userId` with `Guid.TryParse` before calling `FindByIdAsync`. Good catch on the exception surface — `ConvertIdFromString` throwing `FormatException` before EF runs would indeed escape as 500. New integration test `Me_WithNonGuidBearerSubClaim_Returns_401` mints a signed bearer with `sub="not-a-guid"` and pins the 401 contract.

backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
```diff
@@ -0,0 +1,233 @@
...
+    // `HttpOnly=false` on the `__Host-Xsrf-Request` cookie is intentional and
+    // is the entire point of the SPA double-submit pattern (DEC-054): the
+    // framework-managed `__Host-Xsrf` cookie is `HttpOnly=true` and holds the
+    // server-bound secret; this companion is the token-echo channel the SPA
+    // reads and copies into the `X-XSRF-TOKEN` request header. Making this
+    // cookie `HttpOnly=true` would break every antiforgery-gated POST from
+    // the browser. Both cookies carry `__Host-` + `Secure` + `Path=/` per
+    // RFC 6265bis, and antiforgery validation requires the header to match
+    // the server-held token — so an XSS that reads this cookie value still
+    // cannot forge a request because the framework cookie is unreadable.
```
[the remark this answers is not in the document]
leehopper: Addressed in db6de3d — comment rewritten. You're right, the XSS claim conflates two separate threat models. The rewrite now says the pattern defends against CSRF, not XSS: "an injected same-origin script can still issue a request and the browser will attach the framework-managed antiforgery cookie automatically. XSS mitigation is the SPA's problem (CSP, input sanitisation, dependency hygiene)."

backend/tests/RunCoach.Api.Tests/Modules/Identity/AuthControllerIntegrationTests.cs
```diff
@@ -0,0 +1,625 @@
...
+        response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+        response.Content.Headers.ContentType!.MediaType.Should().Be("application/problem+json");
+        using var doc = JsonDocument.Parse(
+            await response.Content.ReadAsStringAsync(TestContext.Current.CancellationToken));
+        doc.RootElement.TryGetProperty("errors", out var errors).Should().BeTrue();
+        var errorKeys = errors.EnumerateObject().Select(p => p.Name).ToArray();
+        var hasPasswordBucket = errorKeys.Any(k => string.Equals(k, "password", StringComparison.OrdinalIgnoreCase));
+        hasPasswordBucket.Should()
+            .BeTrue($"weak-password errors must surface under the DTO-property bucket — got keys: [{string.Join(", ", errorKeys)}]");
+    }
```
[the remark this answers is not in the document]
leehopper: Addressed in db6de3d — renamed to `Register_IdentityPolicyFailure_Returns_400_WithPasswordBucket` and switched the password to `"alllowercase12"` (passes DTO `[MinLength(12)]`, fails Identity's `RequireUppercase` + `RequireDigit` + `RequireNonAlphanumeric`). The test now exercises the `IdentityResult` → `ValidationProblemDetails` mapping path (DEC-052) instead of the `[ApiController]` auto-400 DTO path. The DTO-level rejection is already covered indirectly by `[ApiController]`'s auto-400 and the `RegisterRequestDto` DataAnnotations tests Slice 0's contract tests pin separately.

backend/Dockerfile
```diff
@@ -1,4 +1,43 @@
...
+# Colima 0.10.1 + Docker Desktop-less containerd on an M4 Pro host — all
+# failed 5/5. `--platform=linux/amd64` (Rosetta) also failed (SIGABRT; this
+# Colima profile lacks a working binfmt handler for cross-arch emulation).
+#
+# Durable posture: this Dockerfile is the CI image-build path for x86_64
+# GitHub Actions runners, which are unaffected by #122608. Apple-Silicon
+# contributors use Path B (host-run API + Compose for infra only) per
+# CONTRIBUTING.md and DEC-045. Remove this comment and retry Path A on the
+# first .NET 11 SDK bump.
+FROM mcr.microsoft.com/dotnet/sdk:10.0@sha256:8a90a473da5205a16979de99d2fc20975e922c68304f5c79d564e666dc3982fc AS build
```
leehopper, line 23: **Deferred — pushback on Sonar docker:S8431.**

The tag+digest combination is intentional per DEC-056. Sonar's rule argues the two are redundant; for this project they are complementary:

- **`:10.0`** — human-readable tag so a diff reader can see at a glance which .NET major we're pinned to without resolving hashes.
- **`@sha256:…`** — immutable content pin so Dependabot + supply-chain review can detect and approve digest changes independently of tag bumps.

Dropping the tag loses the human signal; dropping the digest loses the supply-chain guarantee. DEC-056 specifically calls this out as the durable posture until .NET 11 GA (when the SVE2 SIGILL fixes and we can revisit). Not addressing in this PR.

backend/Dockerfile
```diff
@@ -9,9 +48,16 @@ RUN dotnet restore RunCoach.slnx
 COPY . .
 RUN dotnet publish src/RunCoach.Api/RunCoach.Api.csproj -c Release -o /app/publish --no-restore
 
-FROM mcr.microsoft.com/dotnet/aspnet:10.0 AS runtime
+FROM mcr.microsoft.com/dotnet/aspnet:10.0@sha256:55e37c7795bfaf6b9cc5d77c155811d9569f529d86e20647704bc1d7dd9741d4 AS runtime
```
leehopper, line 51: **Deferred — pushback on Sonar docker:S8431.**

Same rationale as [the Dockerfile:23 thread](https://github.com/leehopper/personal-running-coach/pull/50/files). Tag+digest on the aspnet runtime base carries the same DEC-056 justification as the SDK base: human-readable `:10.0` for at-a-glance review + immutable `@sha256:…` for supply-chain pinning.

backend/src/RunCoach.Api/Infrastructure/AntiforgeryBridgeMiddleware.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            HttpContext = ctx,
+            ProblemDetails = new ProblemDetails
+            {
+                Status = StatusCodes.Status400BadRequest,
+                Title = "Antiforgery validation failed.",
+                Type = AntiforgeryProblemType,
+                Detail = "The antiforgery cookie or header was missing or did not match. Call GET /api/v1/auth/xsrf to seed fresh tokens and retry.",
+            },
+        });
```
leehopper, line 53: **💡 [LOW] intent: DEC-055 doc drift — bridge shipped with two rejected alternatives**

**Observation.** DEC-055's *Alternatives considered (and rejected)* section states:

> *"Emit a ProblemDetails body from the bridge middleware via `IProblemDetailsService.TryWriteAsync`. Rejected for consistency — the pre-DEC-055 shape returned a plain 400 with no body (from `BadRequestResult` inside the framework's antiforgery filter), and T02.5's tests pin that contract."*

DEC-055 also rejects the named-class form in favour of an inline delegate (*"Write a named middleware class — rejected for 3 lines; inline is clearer"*). The shipped code is **both** a named static class (`AntiforgeryBridgeMiddleware`) **and** calls `TryWriteAsync` with a populated ProblemDetails body. Integration test Case 5 actively verifies the body, confirming the change was intentional — but DEC-055 was never updated to reflect the supersession.

**Why this matters.** The decision log contradicts the shipped code on two separate points. Future readers who trust the log will assume the current code still matches; a refactor that "restores" the DEC-055 spec would silently change the 4xx contract Case 5 now pins.

**Suggestion.** Add a one-paragraph addendum to DEC-055 documenting that (a) the ProblemDetails body was adopted before merge because the SPA handles every 4xx uniformly — the rationale the Case 5 test comment already captures — and (b) the named-class form was chosen over the inline delegate for testability / future growth (audit logging, richer detail). No code change required; this is a spec-vs-shipped-code drift.

backend/src/RunCoach.Api/Modules/Identity/IdentityResultExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+                _ => IdentityErrorBuckets.General,
+            };
+            controller.ModelState.AddModelError(key, error.Description);
+        }
+
+        return controller.ValidationProblem(controller.ModelState);
+    }
+
+    private static string ToCamelCase(this string value) =>
+        value.Length == 0 ? value : char.ToLowerInvariant(value[0]) + value[1..];
```
leehopper: **💡 [LOW] simplify: `ToCamelCase` helper is redundant — bucket constants already hold lowercased strings**

`IdentityErrorBuckets.Password == "password"` and `IdentityErrorBuckets.Email == "email"` (both already lowercase). The switch arms produce `nameof(RegisterRequestDto.Password).ToCamelCase() == "password"` and `nameof(RegisterRequestDto.Email).ToCamelCase() == "email"` — byte-identical to the bucket constants already matched by the pattern. `ToCamelCase` adds no information.

**Nuance for the refactor.** The `UserName` arm deliberately maps to `email` (not `username`) because `RegisterRequestDto` has no `UserName` field. The correct substitution uses `IdentityErrorBuckets.Email` for **both** the `Email` and `UserName` arms:

```suggestion
            var key = IdentityErrorCodeMapper.Map(error).PropertyName switch
            {
                IdentityErrorBuckets.Password => IdentityErrorBuckets.Password,
                IdentityErrorBuckets.Email => IdentityErrorBuckets.Email,
                IdentityErrorBuckets.UserName => IdentityErrorBuckets.Email,
                _ => IdentityErrorBuckets.General,
            };
            controller.ModelState.AddModelError(key, error.Description);
        }

        return controller.ValidationProblem(controller.ModelState);
    }
```

Then delete the `ToCamelCase` private extension at the bottom of the file. Phase-7 challenger confirmed the strings are byte-identical at 95+ confidence.

backend/src/RunCoach.Api/Infrastructure/AuthCookieNames.cs
```diff
@@ -0,0 +1,46 @@
...
+/// double-submit convention the SPA echoes per DEC-054.
+/// </summary>
+public static class AuthCookieNames
+{
+    /// <summary>
+    /// The Identity application session cookie. Written on successful
+    /// login, rotated on logout, carries the encrypted authentication
+    /// ticket. <c>HttpOnly=true</c>.
+    /// </summary>
+    public const string Session = "__Host-RunCoach";
```
leehopper, line 25: **💡 [LOW] comment: session cookie is cleared on logout, not rotated**

The XML doc says the cookie is *"rotated on logout"* but `Logout` calls `signInManager.SignOutAsync()` which emits an expired `Set-Cookie` (epoch expires / max-age=0) to delete the cookie. `IssueAntiforgeryTokens()` immediately after rotates the **antiforgery** cookies, not the session cookie. In session-cookie terminology, rotation means issuing a new valid ticket; deletion is "cleared" or "expired."

```suggestion
    /// <summary>
    /// The Identity application session cookie. Written on successful
    /// login, cleared (expired) on logout, carries the encrypted
    /// authentication ticket. <c>HttpOnly=true</c>.
    /// </summary>
    public const string Session = "__Host-RunCoach";
```

If calling out that antiforgery tokens are re-issued on logout is useful context, add a separate sentence rather than repurposing "rotated" for deletion.

backend/src/RunCoach.Api/Modules/Identity/IdentityErrorCodeMapper.cs
```diff
@@ -0,0 +1,67 @@
...
+/// <see cref="IdentityErrorBuckets.General"/> / <see cref="IdentityErrorKind.Unknown"/>.
+/// </summary>
+public static class IdentityErrorCodeMapper
+{
+    public static IdentityErrorMapping Map(IdentityError error)
+    {
+        ArgumentNullException.ThrowIfNull(error);
+        return error.Code switch
+        {
+            nameof(IdentityErrorDescriber.PasswordTooShort) =>
```
leehopper: **💡 [LOW] simplify: `IdentityErrorKind` and the `Kind` field are production-dead**

`IdentityErrorCodeMapper.Map` returns `IdentityErrorMapping(PropertyName, Kind)` for all 22 codes. The only production consumer — `IdentityResultExtensions.ToRegistrationActionResult` — reads **only** `.PropertyName`. `Kind` is consumed only by `IdentityErrorCodeMapperTests`. Phase-7 LSP verification confirmed no other `.Kind` reader exists.

**Two paths:**

1. **Delete scaffolding.** Simplify `Map` to return the property-name string directly, drop `IdentityErrorKind.cs` and `IdentityErrorMapping.cs`:

   ```csharp
   public static string MapToPropertyName(IdentityError error) => error.Code switch { ... };
   ```

2. **Mark as intentional.** If `Kind` is scaffolding for upcoming non-registration flows (login, password reset, role management — the mapper covers `Role`/`Conflict`/`Unauthorized` cases reachable only there per DEC-052), add a one-line XML doc to `IdentityErrorKind` explicitly stating "populated for forward-compat; only `.PropertyName` is read today." Otherwise a future reader will reasonably delete it.

Phase-7 contested this (structural-vs-scaffolding judgment); either option closes the signal.

leehopper (commented):
# Deep Review — 7 findings

Full pipeline: 7 Phase-3 agents → 13 findings → Phase 4 verification → Phase 5 validation → Phase 6 reconcile → Phase 7 blind challenge → 7 final findings (1 main, 6 suggestions).

**Overall:** the auth module is carefully defended. Timing-safe login (DEC-053), `__Host-` cookie posture (DEC-050), strict JWT validation (ValidateIssuer/Audience/Lifetime/IssuerSigningKey + RequireSignedTokens + HS256 pinned), antiforgery bridge ordering, `/me` GUID guard, and dev-only stack traces all hold up to scrutiny. Security reviewer returned zero findings above the 70-confidence bar.

**Notable correction from Phase 7:** three findings claiming the `JwtBearerOptionsSetup` name-guard is dead code were all eliminated. The Opus challenger confirmed that `OptionsFactory<T>.Create` performs a runtime `is IConfigureNamedOptions<T>` pattern-match on each registered `IConfigureOptions<T>`, so the named overload and its guard are genuinely reachable. Registering as `IConfigureOptions<T>` while implementing `IConfigureNamedOptions<T>` is the standard ASP.NET Core pattern. **Do not change the registration.**

Full report at `.deep-review/deep-review-report-61b7c1ee.md` (gitignored).

---

🤖 Generated by deep-review (frontier mode). Reviewed up to: 61b7c1ee

backend/tests/RunCoach.Api.Tests/Modules/Identity/AuthControllerIntegrationTests.cs
```diff
@@ -0,0 +1,666 @@
...
+
+        // Act
+        var response = await bearerClient.GetAsync(
+            "/api/v1/auth/me",
+            TestContext.Current.CancellationToken);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        var body = await response.Content.ReadFromJsonAsync<AuthResponseDto>(
+            cancellationToken: TestContext.Current.CancellationToken);
```
leehopper, line 476: **🟠 [HIGH] test: no negative-path coverage for JWT bearer validation flags**

`JwtBearerOptionsSetup` sets six hard-gate validation flags: `ValidateIssuer`, `ValidateAudience`, `ValidateLifetime`, `ValidateIssuerSigningKey`, `RequireSignedTokens`, `ValidAlgorithms=[HS256]`. The bearer branch has two tests — `Me_WithBearerSubClaim_Returns_200` (happy path, Case 10b) and `Me_WithNonGuidBearerSubClaim_Returns_401` (malformed sub). `MintBearerTokenWithRawSub` **always** uses the configured key/issuer/audience/HS256/5-min expiry — a fully valid token. No test exercises a token signed with a wrong key, wrong issuer, wrong audience, past-expiry, or non-HS256.

Bearer is reachable today via `/me` (the happy-path test proves the pipeline is live through `CookieOrBearer`). If any `Validate*` flag were silently flipped during debugging, the handler would accept arbitrary tokens on a production-accessible endpoint and no test would catch it.

**Suggestion.** Add four parameterised cases via a helper like `MintBearerTokenWith(key, issuer, audience, lifetime)`:

| Case | Override | Expected |
|---|---|---|
| wrong key | `signingKey = DifferentHs256Key()` | 401 |
| wrong issuer | `issuer = "https://evil.example"` | 401 |
| wrong audience | `audience = "other-audience"` | 401 |
| expired | `expires = DateTime.UtcNow.AddSeconds(-1)` | 401 |

All four should return `HttpStatusCode.Unauthorized` from `GET /api/v1/auth/me`. Severity is marked high because regression surface is auth-critical; confidence is 57 because the bearer path is dormant per DEC-051 — you can reasonably defer this until the iOS shim lands and move the 401-shape assertions with it.

backend/tests/RunCoach.Api.Tests/Modules/Identity/AuthControllerIntegrationTests.cs
```diff
@@ -0,0 +1,666 @@
...
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        var sessionHeader = GetRawSetCookie(response, SessionCookieName);
+        sessionHeader.Should().NotBeNull($"login must issue {SessionCookieName}");
+        var lowered = sessionHeader!.ToLowerInvariant();
+        lowered.Should().Contain("httponly");
+        lowered.Should().Contain("secure");
+        lowered.Should().Contain("samesite=lax");
+
```
leehopper, line 238: **💡 [LOW] test: no `Path=/` assertion on the session cookie — `__Host-` invariants not pinned**

`Login_HappyPath_Returns_200_AndIssuesSessionCookie` asserts the raw `Set-Cookie` contains `httponly`, `secure`, and `samesite=lax`, but never asserts `path=/` or the absence of `domain=`. Per RFC 6265bis, browsers silently reject `__Host-` prefixed cookies when either invariant fails.

The `__Host-Xsrf-Request` companion cookie is only checked for value non-emptiness via `CookieContainer`, which strips attributes — so its `Secure` / `Path=/` / `HttpOnly=false` contract is never asserted via raw headers either.

**Consequence.** If `cookie.Cookie.Path = "/"` were ever removed from `Program.cs`, auth would break for every browser user but the test suite would pass green.

```suggestion
        lowered.Should().Contain("httponly");
        lowered.Should().Contain("secure");
        lowered.Should().Contain("samesite=lax");
        lowered.Should().Contain("path=/");
        lowered.Should().NotContain("domain=");
```

Add a parallel raw-`Set-Cookie` assertion for `__Host-Xsrf-Request` in `Xsrf_Returns_204_WithRequestTokenCookie` using the existing `GetRawSetCookie` helper — assert `secure`, `path=/`, no `httponly`, no `domain=`.

backend/src/RunCoach.Api/Modules/Identity/AuthController.cs
```diff
@@ -0,0 +1,242 @@
...
+/// </summary>
+[ApiController]
+[Route("api/v1/auth")]
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Major Code Smell", "S6960:Controllers should not have mixed responsibilities", Justification = "Spec §Unit 2 mandates a single AuthController holding the five auth endpoints (xsrf, register, login, me, logout).")]
+public sealed partial class AuthController(
+    UserManager<ApplicationUser> userManager,
+    SignInManager<ApplicationUser> signInManager,
+    IPasswordHasher<ApplicationUser> passwordHasher,
+    IAntiforgery antiforgery,
+    ILogger<AuthController> logger) : ControllerBase
```
leehopper, line 32: **💡 [LOW] convention: AuthController inherits ControllerBase directly**

`backend/CLAUDE.md` line 64 states *"Controllers inherit from `BaseController` (`Common/`)"*. `AuthController` inherits `ControllerBase` with no documented exception.

**Nuance.** `BaseController`'s `[Route("api/v1/[controller]")]` would produce `api/v1/Auth` (PascalCase), but `AuthController` uses the literal lowercase `api/v1/auth` deliberately. Phase-7 also found that `BaseController` has **zero references** anywhere in the codebase — nothing currently inherits from it.

Three options:

1. **Make it the pattern** — have `AuthController` inherit `BaseController` (its own `[Route]` attribute overrides the inherited one, so `api/v1/auth` survives), and note in `BaseController` that literal routes are supported via override.
2. **Update the convention** — add a short exception to `backend/CLAUDE.md` documenting that controllers requiring a literal route may inherit `ControllerBase` directly.
3. **Delete `BaseController`** — if nothing uses it and the convention isn't being enforced, the class is dead weight.

Option 2 is the smallest change and matches current behavior.

backend/src/RunCoach.Api/Modules/Identity/IdentityResultExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+            e.Code is nameof(IdentityErrorDescriber.DuplicateEmail)
+                   or nameof(IdentityErrorDescriber.DuplicateUserName));
+
+        if (hasConflict)
+        {
+            return controller.Problem(
+                type: RegistrationConflictType,
+                title: "The account could not be created.",
+                statusCode: StatusCodes.Status409Conflict);
+        }
```
leehopper, line 50: **💡 [LOW] type: conflict detection re-implements what `IdentityErrorMapping.Kind` already encodes**

Paired with the `IdentityErrorCodeMapper.Kind`-is-dead finding on the mapper file.

The `hasConflict` check hardcodes two codes (`DuplicateEmail`, `DuplicateUserName`), but `IdentityErrorCodeMapper` already classifies each error by `IdentityErrorKind.Conflict` — a richer set that also includes `UserAlreadyHasPassword`, `ConcurrencyFailure`, `LoginAlreadyAssociated`, `UserLockoutNotEnabled`, and the three role codes.

**Current impact:** not triggerable from `CreateAsync` on a fresh registration — Phase-7 verified that none of the non-Duplicate `Conflict` codes surface from that call site. So this is a forward-compat / consistency concern, not a live bug.

**Consistency suggestion:** when/if `simplify-2` is resolved by keeping `Kind` (the "mark as intentional scaffolding" option), also make the conflict check read `.Kind` instead of `error.Code` so there's a single authority for what counts as 409:

```csharp
var mappings = result.Errors.Select(IdentityErrorCodeMapper.Map).ToList();
var hasConflict = mappings.Any(m => m.Kind == IdentityErrorKind.Conflict);
```

If `simplify-2` is resolved by dropping `Kind` entirely (the "delete scaffolding" option), this comment is moot — the hardcoded check becomes the canonical form. Either way these two findings resolve together.

backend/tests/RunCoach.Api.Tests/Modules/Identity/AuthControllerIntegrationTests.cs
```diff
@@ -0,0 +1,666 @@
...
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.NoContent);
+        var sessionClear = GetRawSetCookie(response, SessionCookieName);
+        sessionClear.Should().NotBeNull($"logout must re-issue {SessionCookieName} to clear it");
+        var isCleared =
+            sessionClear!.Contains("expires=thu, 01 jan 1970", StringComparison.OrdinalIgnoreCase) ||
+            sessionClear.Contains("max-age=0", StringComparison.OrdinalIgnoreCase);
+        isCleared.Should().BeTrue(
+            $"cookie-clear must use an epoch expires attribute or max-age=0, got: {sessionClear}");
+    }
```
leehopper, line 447: **💡 [LOW] test (observation, low-value): logout test stops at cookie clear, no follow-up /me**

**Observation.** `Logout_Authenticated_Returns_204_AndClearsSessionCookie` asserts only that the response `Set-Cookie` expires the session cookie; there's no subsequent `GET /me` to confirm the authenticated state is gone.

**Why I'm not suggesting adding one.** ASP.NET Core cookie auth is stateless — the encrypted ticket *is* the session. `signInManager.SignOutAsync()` for cookie auth primarily emits a delete `Set-Cookie` header and does NOT rotate the Identity `SecurityStamp`. If someone swapped it for a bare `Response.Cookies.Delete`, a same-client follow-up `/me` would pass in both cases because the `CookieContainer` drops the expired cookie before sending the next request. The assertion wouldn't catch the threat model it's meant to catch.

**Action:** no action warranted. Recording this here so the gap isn't rediscovered and "fixed" with an ineffective assertion. The real defense for a stolen ticket on another client is the Identity `SecurityStamp` + `SecurityStampValidator` — which isn't exercised by this endpoint and doesn't need to be (logout is local-client only; cross-device revocation belongs to a future "sessions" feature).

leehopper (commented):
# Deep Review — 2 additional notes

Follow-up to the main review posting — two findings from the full pipeline that warrant visibility but weren't in the first batch:

- **type-1:** related to `simplify-2` (also posted). The `IdentityErrorMapping.Kind` inconsistency is real but not currently triggerable.
- **test-3:** observation-only note explaining why a follow-up `/me` after logout is *not* a meaningful test addition given ASP.NET Core cookie-auth statelessness.

🤖 Generated by deep-review.

PR #145, opened 2026-06-05
feat(slice-2b): workout-log persistence foundation

leehopper:
> **Re-opened for proper review.** This is PR2 of slice-2b (originally merged as #138 without the review cycle, then reverted via #144). Same code, now awaiting CodeRabbit + your `/review-pr`. **Auto-merge is OFF — this will not merge until you say so.**

## What & why

**PR2 of the slice-2b 8-PR strategy (Unit 2 — `WorkoutLog` persistence foundation).** Pure persistence: the EF Core entity, prescription snapshot, canonical metric keys, value converters, migration, and repository. **No endpoints, no LLM, no frontend** — those land in later PRs. Leaves `main` green with the entity unreferenced by any endpoint.

Implements spec `docs/specs/16-spec-slice-2b-workout-logging/` Unit 2; decisions **DEC-072** (metrics bag + canonical keys + first value converters) and **DEC-076** (server-authoritative prescription snapshot).

## What's included

- **`WorkoutLog`** EF entity (tenanted): `OccurredOn` + value-converted `Distance`/`Duration`, `CompletionStatus`, freeform `Notes` (no cap), an open `jsonb` `Metrics` bag, typed `jsonb` `Splits`, audit fields, and a nullable prescription snapshot.
- **`WorkoutPrescriptionSnapshot`** = **EF Core 10 optional complex type**, table-split to real columns (DEC-076): `SourcePlanId`/`WeekNumber`/`DayOfWeek` coordinate + frozen prescribed distance/duration/pace. Off-plan ⇒ `Prescription == null`. `WorkoutType` stored **by name** (`text`), reorder-safe.
- **`WorkoutMetricKeys`**: single-sourced canonical keys — a `const` set + `WorkoutMetricKey` enum kept drift-free by a unit test (DEC-072).
- **First EF `ValueConverter`s** — `Distance`↔meters, `Duration`↔ticks, `Pace`↔sec/km — plus a new `Duration` value object.
- **`WorkoutLogRepository`**: create, get-by-id, keyset `read-by-user`, and the `(SourcePlanId, WeekNumber, DayOfWeek)` coordinate query.
- **Migration `AddWorkoutLog`** — `jsonb` columns, FK cascade, table-split nullable prescription columns, keyset index.

## Testing

- **Unit**: metric-key drift guard, `Duration`, value-converter units-preservation.
[...]

PR #147, opened 2026-06-06
feat(slice-2b): workout-log create endpoint with ef-native idempotency

leehopper:
## Slice 2b · PR3 — Workout-log create endpoint

`POST /api/v1/workouts/logs` (Unit 3). Persists a workout log with a **server-authoritative prescription snapshot** (DEC-076) resolved from the runner's active plan, idempotent on a client-supplied `Guid` key. Pure persistence — no LLM on this path (DEC-073). Depends on PR1 (`PlanCalendar`) + PR2 (`WorkoutLog` foundation), both on `main`.

### The idempotency decision (DEC-077, from R-081)

The 2b spec inherited the phrase *"co-transactional Marten `IdempotencyMarker` pattern"* for this endpoint — but that was a carry-over from the Marten-**handler** regenerate flow. `WorkoutLog` is a pure-EF immutable fact with no Marten stream and no handler, so that mechanism is **impossible** here (R-069: the Marten/EF shared-connection co-commit exists only inside `EfCore*Projection` apply paths). A mid-implementation investigation + a deep-research pass (`docs/research/artifacts/batch-29a-…`) eliminated the two alternatives (Wolverine co-commit = impossible; event-sourcing = contradicts DEC-076) and ratified:

- **EF-native idempotency** — an `IdempotencyKey` column + unique index `(UserId, IdempotencyKey)` on `WorkoutLog`, inserted in the **same `SaveChanges`** as the row (atomic by construction). A replayed key catches SQLSTATE `23505` (guarded on `ConstraintName`) and re-reads the prior `WorkoutLogId`; a failed attempt writes nothing, leaving the key reusable. Not a "third store" — one column + one index on a row already being inserted.

The Marten-document `IIdempotencyStore` stays the mechanism for the LLM/handler endpoints (onboarding/regenerate). Full reasoning in **DEC-077**.

### What's in it

- `WorkoutLogsController` (auth + antiforgery + `201`/`400` ProblemDetails) → `IWorkoutLogService` (snapshot resolution via `PlanCalendar.ResolveSlot` over the active plan; off-plan ⇒ null) → `IWorkoutLogRepository.CreateIdempotentAsync`.
- `WorkoutPrescriptionSnapshot.Create` validating factory (fast ≤ slow pace guard) — the deferred **PR2-review follow-up (a)**, so an inverted-bounds snapshot fails at construction, not at read time.
- EF migration `AddWorkoutLogIdempotencyKey`; request/response/split DTOs; regenerated `swagger.json` + RTK/Zod codegen (DEC-066, drift gate green).
- Decision + research: **DEC-077**, R-081 artifact + prompt, research-queue entry.

### Acceptance criteria (`create-endpoint.feature`)

All 9 scenarios covered by integration tests: minimum / rich (notes+metrics+splits) / on-plan server-resolved snapshot (≠ client values) / off-plan no-slot / no active plan / idempotent replay / failed-attempt-then-retry / 401 / antiforgery-reject. Plus 2 unit tests for the snapshot factory guard.

### Verification

- Backend **1199/1199**, frontend **383/383**, `npm run build` clean.
- Pre-push gates green: `codegen-check`, `dotnet-build`, `eval-cache-replay`, `typescript-check`.

> Note: the `docs/specs/16-spec-…` Unit 3 wording + `CreateAsync` doc were corrected to match DEC-077, but `docs/specs/` is gitignored (working-tree-only per repo convention), so those edits aren't in this diff.

## Summary by CodeRabbit

* **New Features**
  * POST /api/v1/workouts/logs to create workout logs with idempotent request handling (safe retries return original record).
  * Accepts rich payloads: distance, duration, completion status, optional notes, JSON metrics, and splits; returns 201 with created log ID and location.

* **Behavior & Validation**
  * Rejects empty idempotency keys and invalid payloads with 400; unauthenticated requests return 401.
  * Server resolves and persists prescription info when available.

* **Documentation**
  * Added design and research docs detailing the idempotency approach.


backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogsController.cs
```diff
@@ -0,0 +1,87 @@
...
+
+/// <summary>
+/// Slice-2b workout-logging endpoints. Every state-changing route is gated by
+/// both <see cref="AuthPolicies.CookieOrBearer"/> and
+/// <see cref="RequireAntiforgeryTokenAttribute"/> per the repo's DEC-055 contract.
+/// </summary>
+[ApiController]
+[Route("api/v1/workouts/logs")]
+[Authorize(Policy = AuthPolicies.CookieOrBearer)]
+public sealed class WorkoutLogsController(IWorkoutLogService service) : ControllerBase
```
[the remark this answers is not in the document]
leehopper: Injected `ILogger<WorkoutLogsController>` and added structured logging on both rejection paths via the repo's source-generated `LoggerMessage` pattern — direct `Log*` calls fail CA1848/CA1873 here. `LogCreateRejectedMissingUser` (Warning) on the missing/malformed claim path; `LogCreateRejectedInvalid` (Information, `{UserId}`/`{IdempotencyKey}` + exception) on the 400 domain-validation path. Kept `service` as the primary-ctor parameter (matching `OnboardingController`) rather than introducing a `_service` field. Fixed in 6ed1d89.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogsController.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// idempotency on the request's idempotency key (DEC-077). A replayed key
+    /// returns the original log id without creating a duplicate.
+    /// </summary>
+    /// <param name="request">The create-workout-log request body.</param>
+    /// <param name="ct">Cancellation token.</param>
+    /// <returns>201 Created with the new (or replayed) log id; 400 on an invalid request.</returns>
+    [HttpPost]
+    [RequireAntiforgeryToken]
+    [ProducesResponseType(typeof(CreateWorkoutLogResponseDto), StatusCodes.Status201Created)]
+    [ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status400BadRequest)]
```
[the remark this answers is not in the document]
leehopper: Added `[ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status401Unauthorized)]` to the POST action and regenerated `backend/openapi/swagger.json` — the endpoint now declares 201/400/401. Fixed in 6ed1d89.

backend/tests/RunCoach.Api.Tests/Modules/Training/Workouts/WorkoutLogsControllerCreateIntegrationTests.cs
```diff
@@ -0,0 +1,474 @@
...
+        var secondBody = await second.Content.ReadFromJsonAsync<CreateWorkoutLogResponseDto>(
+            cancellationToken: ct);
+
+        // Assert
+        first.StatusCode.Should().Be(HttpStatusCode.Created);
+        ((int)second.StatusCode).Should().BeOneOf(
+            StatusCodes.Status200OK, StatusCodes.Status201Created);
+        firstBody.Should().NotBeNull();
+        secondBody.Should().NotBeNull();
+        secondBody!.WorkoutLogId.Should().Be(firstBody!.WorkoutLogId, because: "a replay returns the original id");
```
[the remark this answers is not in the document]
leehopper: Pinned the replay to `second.StatusCode.Should().Be(HttpStatusCode.Created)` (removed `BeOneOf`) and added `secondBody.Should().BeEquivalentTo(firstBody)` for full deep-equality. The response DTO carries only `WorkoutLogId` (no audit fields), so this is the complete body contract. Fixed in 6ed1d89.

backend/tests/RunCoach.Api.Tests/Modules/Training/Workouts/WorkoutLogsControllerCreateIntegrationTests.cs
```diff
@@ -0,0 +1,474 @@
...
+        var bad = MinimalRequest(Week2Day4, key) with { DistanceMeters = -1.0 };
+        var failed = await PostLogAsync(client, token, bad, ct);
+
+        // Retry the same key with a valid body.
+        var retry = await PostLogAsync(client, token, MinimalRequest(Week2Day4, key), ct);
+
+        // Assert
+        ((int)failed.StatusCode).Should().BeGreaterThanOrEqualTo(
+            400, because: "the invalid first attempt must not succeed");
+        retry.StatusCode.Should().Be(HttpStatusCode.Created, because: "the same key with a valid body must still create the log");
```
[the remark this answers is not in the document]
leehopper: Tightened to `failed.StatusCode.Should().Be(HttpStatusCode.BadRequest)` plus a `ContentType.MediaType == "application/problem+json"` assertion, matching the sibling `Create_AuthenticatedNoAntiforgery_Rejected` test. A 500 regression now fails the suite. Fixed in 6ed1d89.

backend/tests/RunCoach.Api.Tests/Modules/Training/Workouts/WorkoutPrescriptionSnapshotTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            dayOfWeek: 4,
+            workoutType: WorkoutType.Tempo,
+            prescribedDistance: Distance.FromKilometers(10.0),
+            prescribedDuration: Duration.FromMinutes(50.0),
+            prescribedPaceFast: Pace.FromSecondsPerKm(280.0),
+            prescribedPaceSlow: Pace.FromSecondsPerKm(330.0));
+
+        // Assert — the computed PaceRange view is readable without throwing.
+        actual.PrescribedPace.Fast.SecondsPerKm.Should().Be(280.0);
+        actual.PrescribedPace.Slow.SecondsPerKm.Should().Be(330.0);
```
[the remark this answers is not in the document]
leehopper: Added `Create_WithEqualPaceBounds_ExposesPrescribedPaceRange`, asserting equal fast/slow bounds (300 s/km) construct without throwing and expose the single pace. The guard uses a strict `IsSlowerThan` (`>`), so equality is permitted — a mistaken strict-`<` guard would now regress here. Fixed in 6ed1d89.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogsController.cs
```diff
@@ -0,0 +1,103 @@
...
+            // Domain-invariant violations from value-object construction (negative
+            // distance/duration, invalid split) surface as a 400 ProblemDetails
+            // rather than an unhandled 500.
+            LogCreateRejectedInvalid(logger, userId, request.IdempotencyKey, ex);
+            return Problem(
+                type: InvalidLogType,
+                title: "The workout log request is invalid.",
+                detail: ex.Message,
+                statusCode: StatusCodes.Status400BadRequest);
+        }
```
leehopper, line 84: **💡 [LOW] Broad catch(ArgumentException) maps server-side plan-data faults to a client 400 and leaks internal detail**

`CreateAsync` awaits `ResolvePrescriptionAsync` **before** any client value object is built, and that path throws `ArgumentException` from *server-side* data, not client input:
- `PlanCalendar.ResolveSlot` throws when the stored plan `PlanStartDate` is not a Sunday;
- `WorkoutPrescriptionSnapshot.Create` throws on inverted stored pace bounds;
- `Pace.FromSecondsPerKm` throws `ArgumentOutOfRangeException` (an `ArgumentException`) on a zero/negative stored target pace.

The pace values originate from unvalidated LLM `WorkoutOutput` (no ordering/positivity invariant in plan generation). The client sends no prescription data (DEC-076), yet these faults are returned as `400 "The workout log request is invalid"` with `detail: ex.Message`, exposing internal plan state (`PlanStartDate`, prescribed pace s/km). A 400 is the wrong status here (should be 500), and leaking `ex.Message` violates the CLAUDE.md/REVIEW.md rule against exposing internal exception detail.

**Suggested fix:** Wrap only client value-object construction (`Distance`/`Duration`/splits) in the `try` and return 400 for those; let prescription-resolution faults propagate to the global handler as a 500 (logged with context). Alternatively, treat an inverted-bounds / invalid prescription as off-plan (`return null`) in `ResolvePrescriptionAsync` so a legitimate run still logs.

_Confidence: 80% · deep-review (corroborated by cross-file-impact)_

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogService.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        var prescription = await ResolvePrescriptionAsync(userId, request.OccurredOn, ct).ConfigureAwait(false);
+        var now = _timeProvider.GetUtcNow();
+
+        var log = new WorkoutLog
+        {
+            WorkoutLogId = Guid.NewGuid(),
+            UserId = userId,
+            TenantId = userId.ToString(),
+            IdempotencyKey = request.IdempotencyKey,
```
leehopper, line 44: **💡 [LOW] Guid.Empty accepted as idempotency key — a second distinct workout is silently swallowed as a replay**

`[JsonRequired]` enforces *presence* of `IdempotencyKey`, not a non-default value, so a client can explicitly send the all-zeros UUID `00000000-0000-0000-0000-000000000000` (a common uninitialised-value mistake). The unique index is `(UserId, IdempotencyKey)`: the first `Guid.Empty` request creates a row, and a genuinely different second workout from the same user with `Guid.Empty` trips the `23505` conflict and `CreateIdempotentAsync` returns the **first** log id as a replay — silent data loss with a 201.

The sibling `PlanRenderingController.Regenerate` already rejects `request.IdempotencyKey == Guid.Empty` with a 400, so this endpoint is missing a guard the team applies elsewhere for the identical concern.

**Suggested fix:** Reject `Guid.Empty` at the API boundary (controller or `CreateAsync`) with a 400 ProblemDetails, mirroring `PlanRenderingController`.

_Confidence: 35% · deep-review (validator-contested; survived blind challenge at 96%)_

backend/src/RunCoach.Api/Modules/Training/Workouts/CreateWorkoutLogRequestDto.cs
```diff
@@ -0,0 +1,37 @@
...
+/// <param name="Splits">Optional typed per-lap splits. Null when none.</param>
+public sealed record CreateWorkoutLogRequestDto(
+    [property: JsonRequired] Guid IdempotencyKey,
+    [property: JsonRequired] DateOnly OccurredOn,
+    [property: JsonRequired] double DistanceMeters,
+    [property: JsonRequired] double DurationSeconds,
+    [property: JsonRequired] CompletionStatus CompletionStatus,
+    string? Notes,
+    IReadOnlyDictionary<string, JsonElement>? Metrics,
+    IReadOnlyList<WorkoutLogSplitDto>? Splits);
```
leehopper, line 37: **💡 [LOW] No server-side size cap on Notes / Metrics / Splits persisted verbatim**

`Notes` (text), `Metrics` (jsonb, re-serialized verbatim) and `Splits` (jsonb array) are persisted with no server-side length or element-count validation (the DTO doc explicitly states `Notes` has no cap), and the project has no rate limiting. An authenticated user can write near-body-limit rows (~28.6MB Kestrel default) on every request — storage/cost amplification, bounded by auth + the Kestrel limit (not a breach).

**Suggested fix:** Add explicit bounds at the boundary — `[StringLength]`/`[MaxLength]` on `Notes`, a max element count (and per-value size) on `Metrics`, and a max count on `Splits`, returning 400 ProblemDetails when exceeded; optionally lower the per-endpoint `RequestSizeLimit`.

_Confidence: 35% · deep-review (validator-contested; survived blind challenge at 95%)_

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogRepository.cs
```diff
@@ -23,6 +24,37 @@ public async Task CreateAsync(WorkoutLog log, CancellationToken ct)
...
+        {
+            // Replay: this (user, idempotency key) already produced a log. The failed
+            // SaveChanges auto-rolled-back its own implicit transaction, so the
+            // context's connection is clean for a fresh read; AsNoTracking bypasses
+            // the lingering Added entity (DEC-077 / R-081).
+            var existingId = await _db.WorkoutLogs
+                .AsNoTracking()
+                .Where(w => w.UserId == log.UserId && w.IdempotencyKey == log.IdempotencyKey)
+                .Select(w => w.WorkoutLogId)
+                .SingleAsync(ct);
```
leehopper, line 52: **🟡 [MEDIUM] No test that idempotency keys are scoped per-user (cross-tenant boundary unverified)**

Idempotency uniqueness is the composite index on `(UserId, IdempotencyKey)` and the catch-and-reread query filters on both `UserId` AND `IdempotencyKey` — correct in production. But every replay/idempotency test uses a single user; no test verifies that two **different** users posting the **same** key each get a distinct `WorkoutLogId`. The per-user scoping is the tenant-isolation boundary, and it is exactly the property left unverified — a future narrowing of the scope to the key alone would be caught by no test.

**Suggested fix:** Add an integration test: register users A and B, POST the same `idempotencyKey` from both, assert two 201s with **distinct** ids and that each user's `GetByUser` returns only their own log.

_Confidence: 67% · deep-review (improvement suggestion)_

backend/src/RunCoach.Api/Modules/Training/Workouts/CreateWorkoutLogResponseDto.cs
```diff
@@ -0,0 +1,11 @@
+namespace RunCoach.Api.Modules.Training.Workouts;
+
+/// <summary>
+/// Response body for <c>POST /api/v1/workouts/logs</c> (slice-2b Unit 3). Carries
+/// the newly-created log id so the frontend can correlate the pessimistic create
+/// and invalidate its <c>WorkoutLog</c> RTK-query tag. Recorded on the
+/// idempotency marker so a replayed key returns the byte-identical id without
+/// creating a second row.
```
leehopper: **💡 [LOW] Response DTO comment references a non-existent “idempotency marker”**

The XML doc says the id is “Recorded on the idempotency marker so a replayed key returns the byte-identical id.” DEC-077 explicitly **rejected** any marker: the key is a column on the `WorkoutLog` row (unique index on `(UserId, IdempotencyKey)`), and on replay the prior id is re-read straight from the row by `CreateIdempotentAsync`. No marker entity exists on this path (an `IdempotencyMarker` exists only for the Marten/Wolverine handler endpoints), so the comment describes a discarded design and misleads the reader.

**Suggested fix:** Reword to the EF-native design, e.g. “re-read from the `WorkoutLog` row via its unique `(UserId, IdempotencyKey)` on a replayed key (DEC-077); no separate marker.”

_Confidence: 75% · deep-review (improvement suggestion)_

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutPrescriptionSnapshot.cs
```diff
@@ -49,4 +49,49 @@ public sealed record WorkoutPrescriptionSnapshot
...
     public PaceRange PrescribedPace => new(PrescribedPaceFast, PrescribedPaceSlow);
+
+    /// <summary>
+    /// Creates a snapshot, validating the prescribed pace bounds are ordered
+    /// (fast no slower than slow) at construction so the computed
+    /// <see cref="PrescribedPace"/> view can never throw at read time. The
+    /// server-authoritative create path (PR3) sources already-ordered plan paces;
+    /// this factory makes a future inverted-bounds caller fail loudly at the
+    /// construction site rather than silently persisting a snapshot that throws
+    /// only when later read (deferred PR2-review follow-up).
```
leehopper: **💡 [LOW] Create factory XML doc carries PR-process narrative**

The `Create` XML doc references the create path as `(PR3)` and closes with `(deferred PR2-review follow-up)` — PR-process / review-history narrative. The project comment convention permits only Spec/DEC/R-NNN symbolic refs and forbids planning / chain-of-thought narrative. The load-bearing “why” (validate ordered bounds so the computed `PrescribedPace` view never throws) is already stated and should remain.

**Suggested fix:** Drop `(PR3)` and `(deferred PR2-review follow-up)`; keep the behavioral why and cite DEC-076/DEC-077 if a symbolic ref is wanted.

_Confidence: 75% · deep-review (improvement suggestion)_

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogService.cs
```diff
@@ -0,0 +1,145 @@
...
+        var plan = await session.LoadAsync<PlanProjectionDto>(planId, ct).ConfigureAwait(false);
+        if (plan?.Macro is null)
+        {
+            return null;
+        }
+
+        if (PlanCalendar.ResolveSlot(plan.PlanStartDate, occurredOn, plan.Macro.TotalWeeks) is not { } slot)
+        {
+            return null;
+        }
```
leehopper, line 152: **💡 [LOW] Untested off-plan branches: null Macro and date-outside-plan-range (ResolveSlot null)**

`ResolvePrescriptionAsync` has five early-return-null branches. Integration tests cover the no-active-plan and no-workout-at-resolved-day branches, but two are unexercised: `plan.Macro is null` (line 111) and `PlanCalendar.ResolveSlot` returning null because `OccurredOn` is before `PlanStartDate` or beyond `TotalWeeks` (line 116). Both should yield an off-plan (null prescription) log. (`PlanCalendarTests` covers `ResolveSlot`-null only in isolation, not through the service.)

**Suggested fix:** Add a test that seeds an active plan but logs a date before `PlanStartDate` or past the final week (`ResolveSlot` → null), asserting 201 + null `Prescription`; optionally one with a null `Macro`.

_Confidence: 57% · deep-review (improvement suggestion)_

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutPrescriptionSnapshot.cs
```diff
@@ -49,4 +49,49 @@ public sealed record WorkoutPrescriptionSnapshot
...
+    /// </summary>
+    public static WorkoutPrescriptionSnapshot Create(
+        Guid sourcePlanId,
+        int weekNumber,
+        int dayOfWeek,
+        WorkoutType workoutType,
+        Distance prescribedDistance,
+        Duration prescribedDuration,
+        Pace prescribedPaceFast,
+        Pace prescribedPaceSlow)
```
leehopper, line 70: **💡 [LOW] Create guard is bypassable via public init setters (latent defense-in-depth gap)**

`Create`’s whole purpose is to guarantee fast≤slow so the computed `PrescribedPace` view (`new PaceRange(fast, slow)`, which throws on inverted bounds) never throws at read time. But the record exposes public `{ get; init; }` setters on every property, so `new WorkoutPrescriptionSnapshot { PrescribedPaceFast = slow, PrescribedPaceSlow = fast }` compiles and skips the guard — the invariant is enforced by convention, not structure. The sibling `PaceRange` uses get-only properties to prevent exactly this.

**Caveat (why low / contested):** No production path uses the bypass (only `Create`, with ordered plan paces), and as an EF-mapped complex type EF needs the init setters to materialize; the init-setter design predates this PR (#145). This is a latent gap, not a reachable defect. If you want the structural guarantee, validate inside the record body / a private ctor so every construction path runs the check.

_Confidence: 20% · deep-review (blind-challenge contested at 55%)_

leehopper (commented):
### 🔎 Deep Review — Slice 2b workout-log create endpoint

Frontier (all-Opus) multi-agent review with blind-challenge verification across bugs, security, cross-file impact, tests, conventions, type design, and simplification.

**8 findings — 0 blocking.** The implementation is clean and faithful to DEC-077; `code-simplifier` found nothing and the test suite is thorough (9 integration + 3 unit). Findings are edge cases and polish.

**Action items (3):**
- 💡 **F-01** `WorkoutLogsController` — broad `catch(ArgumentException)` reports server-side plan-data faults as a client 400 and leaks internal detail *(most actionable)*
- 💡 **F-02** `WorkoutLogService` — `Guid.Empty` accepted as idempotency key → 2nd distinct workout silently swallowed as a replay
- 💡 **F-03** `CreateWorkoutLogRequestDto` — no server-side size cap on Notes/Metrics/Splits

**Improvement suggestions (5):**
- 🟡 **S-01** per-user idempotency scoping has no regression test (cross-tenant boundary unverified)
- 💡 **S-02** response DTO comment references a non-existent “idempotency marker” (DEC-077 rejected markers)
- 💡 **S-03** `Create` factory XML doc carries `(PR3)` / `(deferred PR2-review follow-up)` process narrative
- 💡 **S-04** off-plan branches (null `Macro`, date-outside-range) untested
- 💡 **S-05** `WorkoutPrescriptionSnapshot.Create` guard bypassable via public init setters (latent; EF needs the setters)

Verified clean: auth, IDOR/multi-tenancy, SQL injection, secrets, and DEC-077 intent fidelity.

_Frontier mode · validators Sonnet · 9 raw → 8 final · all 8 survived blind challenge (1 contested)._

---
Generated by deep-review | Reviewed up to: 6ed1d89ee32d479b17c5b6918eed48247e69642b

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogsController.cs
leehopper: Fixed in 8393103. `ResolvePrescriptionAsync` now wraps slot-resolution + snapshot construction (the `ResolveSlot` / `WorkoutPrescriptionSnapshot.Create` / `Pace.FromSecondsPerKm` calls — all of which throw `ArgumentException` from *server-side* plan data) in a `catch (ArgumentException)` that logs a warning and returns null, so the run logs off-plan instead of the fault reaching the controller's client-input catch. That catch now only ever fires for client value-object construction (negative distance/duration/splits), so the 400 + `ex.Message` it returns describes the client's own input, never internal plan state (`PlanStartDate`, prescribed pace s/km). Chose off-plan-with-warning over a 500 so a malformed LLM-generated plan can't block a legitimate run from logging.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogService.cs
leehopper: Fixed in 8393103. `CreateLog` now rejects `IdempotencyKey == Guid.Empty` with a 400 ProblemDetails (`invalid-idempotency-key`) at the boundary, mirroring `PlanRenderingController.Regenerate`. Added `Create_WithEmptyIdempotencyKey_Returns400_NoRecordCreated` asserting the 400 + `application/problem+json` + no row persisted.

backend/src/RunCoach.Api/Modules/Training/Workouts/CreateWorkoutLogRequestDto.cs
leehopper: Not adding size caps in this PR. The exposure is bounded by auth + Kestrel's ~28.6MB body limit, and per the slice-2b spec this is a single-user MVP-0 with no rate limiting — the threat model (an authenticated user inflating their own storage) is the developer themselves. Concrete `[StringLength]`/`[MaxLength]` values are an unspecified product decision, and adding DataAnnotations here would also shift the OpenAPI `maxLength`/`maxItems` contract (codegen-drift gate). The structural fix — rate limiting plus spec'd input bounds — is a cross-cutting concern larger than this endpoint; tracking it as backlog rather than landing arbitrary caps as a band-aid.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogRepository.cs
leehopper: Fixed in 8393103. Added `Create_SameKeyDifferentUsers_EachGetsDistinctLog`: registers two users, POSTs the same idempotency key from both, and asserts two 201s with distinct `WorkoutLogId`s plus each user's `GetByUser` returning only their own log. The composite `(UserId, IdempotencyKey)` scoping — and any future narrowing to the key alone — is now regression-covered.

backend/src/RunCoach.Api/Modules/Training/Workouts/CreateWorkoutLogResponseDto.cs
leehopper: Fixed in 8393103. Reworded the doc to the EF-native design: on a replayed key the prior id is re-read straight from the `WorkoutLog` row via its unique `(UserId, IdempotencyKey)` index (DEC-077); there is no separate idempotency marker on this path.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutPrescriptionSnapshot.cs
leehopper: Fixed in 8393103. Dropped `(PR3)` and `(deferred PR2-review follow-up)` from the `Create` XML doc; kept the behavioral why (validate ordered bounds so the computed `PrescribedPace` view never throws at read time) and cited DEC-076.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogService.cs
leehopper: Fixed in 8393103. Added `Create_OnDateBeforePlanStart_PersistsOffPlan`: seeds an active plan, logs a run dated the Sunday before `PlanStartDate` so `ResolveSlot` returns null *through the service*, and asserts 201 + null `Prescription`. That exercises the date-outside-range off-plan branch (line 116) end-to-end. Left the `plan.Macro is null` branch unexercised — `PlanProjectionDto.Macro` is non-null on every seeded/generated plan, so a null-Macro fixture would assert an unreachable state; the branch stays as a cheap defensive guard.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutPrescriptionSnapshot.cs
leehopper: Not changing the init setters. As the finding's own caveat notes, EF maps `WorkoutPrescriptionSnapshot` as an optional complex type and needs the `{ get; init; }` setters to materialize it, no production path uses the bypass (only `Create`, with ordered plan paces), and the design predates this PR (#145). A private-ctor / record-body validation would break EF materialization for a non-reachable defect. Additionally, after 8393103 the `Create` guard's throw is caught in `ResolvePrescriptionAsync` and downgraded to an off-plan log, so even a hypothetical inverted construction can't surface as an error to a caller. Leaving the structural guarantee to `PaceRange` (get-only) as before.

PR #156, opened 2026-06-07
feat(slice-2b): workout-log /log route and logging form (PR6b)

leehopper:
## Slice 2b — PR6b (`/log` route + logging form)

PR6b of the [8-PR slice-2b train](docs/specs/16-spec-slice-2b-workout-logging/pr-strategy.md). 6 of 8 already merged (PR1/PR6a/PR2/PR3/PR4/PR5); this is **Unit 6 — the frontend `/log` route + form**. PR7 (history list + E2E) follows.

### What this delivers
- **`/log` route** (auth + onboarding guarded, mirrors the home route's guard) — a mobile-first single-column logging form.
- **Flexible RHF + Zod form** (DEC-075): core fields (date, distance, duration, completion, notes) always visible; optional metrics (RPE, avg/max HR, elevation gain) behind an always-mounted **"More details" `Collapsible`** (default closed). `shouldUnregister: false` so a filled-then-collapsed metric still submits.
- **Schema derived from the generated request body** via `.pick().extend()` (DEC-075, first use in the repo): `occurredOn` + `completionStatus` stay drift-safe against the backend; the UI uses km / decimal-minutes mapped to meters/seconds on submit. Optional numerics resolve **empty → `undefined` (never `0`/`NaN`)**; distance + duration are required for `Complete`/`Partial`, optional for `Skipped` (→ `0` on the wire).
- **Pessimistic create** with a stable `idempotencyKey` (`useMemo`, DEC-077) → `createWorkoutLog` mutation invalidates a new `WorkoutLog` cache tag → **`Workout logged` toast** → navigate home.
- **Today-card "Log run" action** — a link to `/log` (`OccurredOn` defaults to today; the server resolves the prescription snapshot from the date).
- **Shared `metric-meta` map** mirroring backend `WorkoutMetricKeys` so form/history labels can't drift; `CompletionStatus` const/type + drift guard added to the codegen barrel.
- **`FormField` gains a backward-compatible 3rd `TTransformedValues` generic** so the input≠output (3-generic) form can use the shadcn `Form` stack; existing single-generic forms are unaffected.

### Gherkin scenarios covered
All of `frontend-log-route-and-form.feature` for the form (route renders core fields + collapsed metrics; missing-required inline `role="alert"`; minimum payload submits; empty optional numeric → `undefined`; filled-then-collapsed still submits; today-card Log action → `/log` with `OccurredOn` = today). The today-card **completed-state** scenario is **deferred to PR7** (per decision — it needs the `WorkoutLog` query consumer that PR7 builds).

### Verification
- `npm run build` (tsc + vite) ✅ · `npm run test` **409 passed** ✅ · `npm run lint` 0 errors ✅ · codegen drift gate ✅
- Multi-agent adversarial review: 20 findings → 19 refuted, 1 confirmed (positive/test-quality), 0 real defects.

### Open-question resolutions (Unit 6)
- **Skipped + zero distance/duration:** allowed (blank → `0`), required-`>0` only for Complete/Partial.
- **Off-plan UI entry:** deferred — `/log` already persists off-plan runs (server resolves a null snapshot from the date).

### Follow-ups found
- None new. (Today-card completed-state → PR7 by design.)

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated workout logging page with full form (date, distance, duration, completion status, notes) and save flow.
  * Added "Log run" quick-action on the daily plan card and a collapsible "More details" section for extra metrics (RPE, HR, elevation).
  * Exposed a create-workout API hook for submitting logs.

* **Bug Fixes / UX**
  * Client-side validation, deterministic idempotent submits, success toast and navigation on save; graceful error reporting.

* **Tests**
  * Comprehensive unit and integration tests covering form behavior, submission, retries, and UI states.

### DIFF
--- a/frontend/src/app/api/generated/index.ts
+++ b/frontend/src/app/api/generated/index.ts
@@ -33,6 +33,7 @@ import { z } from 'zod'
 
 import { PostApiV1AuthRegisterBody } from './zod/auth/auth'
 import { PostApiV1OnboardingTurnsResponse } from './zod/onboarding/onboarding'
+import { PostApiV1WorkoutsLogsBody } from './zod/workout-logs/workout-logs'
 
 // Auth schemas — migrated piecewise. RegisterRequest is the first; T03.2
 // extends to OnboardingProgressDto + SuggestedInputType.
@@ -51,6 +52,17 @@ export type OnboardingProgressDto = z.infer<typeof onboardingProgressSchema>
 
 export const suggestedInputTypeSchema = PostApiV1OnboardingTurnsResponse.shape.suggestedInputType
 export type SuggestedInputType = z.infer<typeof suggestedInputTypeSchema>
+
+// Workout-log schemas (slice-2b). The create-request body is the drift seam
+// the `/log` form derives from via `.pick().extend()` (DEC-075): the form keeps
+// `occurredOn`'s ISO-date format and `completionStatus`'s `0|1|2` enum honest
+// against the backend contract, and a backend rename surfaces here as a TS
+// error during `npm run codegen:check`.
+export const createWorkoutLogRequestSchema = PostApiV1WorkoutsLogsBody
+export type CreateWorkoutLogRequest = z.infer<typeof createWorkoutLogRequestSchema>
+
+export const completionStatusSchema = createWorkoutLogRequestSchema.shape.completionStatus
+export type CompletionStatus = z.infer<typeof completionStatusSchema>
 
 /**
  * `as const` enum-shaped object paired with {@link SuggestedInputType}
@@ -87,3 +99,26 @@ type _SuggestedInputTypeExhaustive = _SuggestedInputTypeUnionMinusConst extends 
 // Exported to satisfy `noUnusedLocals`; the assignment fails if a new backend
 // variant widens `SuggestedInputType` without a matching entry in the const above.
 export const _suggestedInputTypeExhaustivenessGuard: _SuggestedInputTypeExhaustive = true
+
+/**
+ * `as const` enum-shaped object paired with {@link CompletionStatus} so the
+ * logging form and history surfaces read `CompletionStatus.Complete` instead of
+ * the magic integer `0`. Values mirror the C# `CompletionStatus` enum on the
+ * wire (no `JsonStringEnumConverter`); the Zod union above enforces the same
+ * `0 | 1 | 2` set at the validation seam. Kept in lockstep with the backend by
+ * the same bidirectional guard used for {@link SuggestedInputType}.
+ */
+export const CompletionStatus = {
+  Complete: 0,
+  Partial: 1,
+  Skipped: 2,
+} as const satisfies Record<string, CompletionStatus>
+
+// Exhaustiveness guard — union ⊆ const direction (mirrors the SuggestedInputType
+// guard above): a new backend `completionStatus` literal that widens the
+// Zod-inferred union without a matching const entry resolves this to `never`,
+// failing the build.
+type _CompletionStatusConstMembers = (typeof CompletionStatus)[keyof typeof CompletionStatus]
+type _CompletionStatusUnionMinusConst = Exclude<CompletionStatus, _CompletionStatusConstMembers>
+type _CompletionStatusExhaustive = _CompletionStatusUnionMinusConst extends never ? true : never
+export const _completionStatusExhaustivenessGuard: _CompletionStatusExhaustive = true
--- a/frontend/src/app/api/generated/index.ts
+++ b/frontend/src/app/api/generated/index.ts
@@ -61,6 +61,12 @@ export type SuggestedInputType = z.infer<typeof suggestedInputTypeSchema>
 export const createWorkoutLogRequestSchema = PostApiV1WorkoutsLogsBody
 export type CreateWorkoutLogRequest = z.infer<typeof createWorkoutLogRequestSchema>
 
+// The create response has no Zod schema to derive from — Swashbuckle emits no
+// 201 body schema, so the generated Zod response is `z.void()`. Re-export the
+// generated RTK *type* (type-only, so the rtk module's runtime self-injection is
+// not pulled in) so the response wire shape stays generated, not hand-mirrored.
+export type { CreateWorkoutLogResponseDto } from './rtk/api'
+
 export const completionStatusSchema = createWorkoutLogRequestSchema.shape.completionStatus
 export type CompletionStatus = z.infer<typeof completionStatusSchema>
 
--- a/frontend/src/app/api/generated/index.ts
+++ b/frontend/src/app/api/generated/index.ts
@@ -17,9 +17,10 @@
  * schemas. The named-component extractions below pull the sub-schemas via
  * Zod's `.shape` accessor so a backend rename of (say)
  * `OnboardingProgressDto.completedTopics` → `OnboardingProgressDto.completed`
- * surfaces here as a TypeScript error during `npm run codegen:check`. That
- * regression class is exactly what Slice 1 bug #4 shipped undetected and
- * what DEC-066 exists to prevent.
+ * is caught by `codegen:check` as a generated-output drift (git-diff) failure;
+ * the resulting broken `.shape` access then surfaces as a TypeScript error
+ * under `npm run build`. That regression class is exactly what Slice 1 bug #4
+ * shipped undetected and what DEC-066 exists to prevent.
  *
  * Drift is caught bidirectionally for `SuggestedInputType`:
  *   - const ⊆ union: enforced by `satisfies Record<string, SuggestedInputType>`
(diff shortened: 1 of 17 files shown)


frontend/src/app/api/workout-log.api.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Response shape returned by `POST /api/v1/workouts/logs`. Mirrors
+ * `RunCoach.Api.Modules.Training.Workouts.CreateWorkoutLogResponseDto`. The
+ * server replies `201 Created` with the persisted log's id; the generated Zod
+ * response schema is `z.void()` (Swashbuckle emits no 201 body schema), so this
+ * hand-mirrored type is the source of truth for the mutation's response generic.
+ */
+export interface CreateWorkoutLogResponseDto {
+  workoutLogId: string
+}
```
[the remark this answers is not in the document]
leehopper: Good catch — applied, with one correction: `CreateWorkoutLogResponseDto` is exported from `generated/rtk/api.ts`, not the `~/api/generated` barrel, so importing it directly from the barrel as proposed wouldn't have compiled. Fixed it the convention-aligned way (commit b4e8375): re-exported the generated RTK type through the barrel (`export type { CreateWorkoutLogResponseDto } from './rtk/api'` — type-only, so the rtk module's runtime self-injection isn't pulled in) and imported it in `workout-log.api.ts`, removing the hand-mirrored interface. Response wire shape is now generated/drift-safe (DEC-066).

frontend/src/app/modules/logging/pages/log.page.tsx
```diff
@@ -0,0 +1,66 @@
...
+  })
+
+  const onSubmit = async (values: WorkoutLogFormValues): Promise<void> => {
+    setFormAlert(null)
+    try {
+      await createWorkoutLog(toCreateWorkoutLogRequest(values, idempotencyKey)).unwrap()
+      toast.success('Workout logged')
+      navigate('/', { replace: true })
+    } catch {
+      setFormAlert('We could not save your workout. Please try again in a moment.')
```
leehopper: **💡 [LOW] Bare catch swallows submit error with no logging or telemetry**

The `onSubmit` catch has no error binding and maps every failure class — a network error, a 400 on a malformed `metrics` bag, a 500, an unexpected `TypeError` — to one generic message. Because the `.unwrap()` rejection is awaited inside `try/catch`, it is a *handled* rejection: it never fires the window `unhandledrejection` event (so `useGlobalErrorReporter` is bypassed) and it is not a render-phase throw (so `AppErrorBoundary` is bypassed). Both pieces of the app's error infrastructure are structurally unreachable from here, leaving zero diagnostic signal when a user reports repeated "could not save" failures.

Note: the sibling `regenerate-plan-dialog.component.tsx` uses the same bare-catch pattern, so this may be an accepted convention rather than an oversight — worth a deliberate call either way.

**Suggested fix:** Bind the error (`catch (error)`) and forward it to the app's client error reporter (or at minimum `console.error`) before showing the user-facing alert, so backend failures stay observable.

---
*Confidence: 75% | deep-review*

frontend/src/app/modules/logging/metric-meta.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Frontend metric-meta map (DEC-072 / DEC-075). The single place UI labels +
+ * units for the open `WorkoutLog.metrics` bag live, so the log form (PR6b) and
+ * the history list (PR7) render identical wording and cannot drift.
+ *
+ * Keys are the canonical lower-camel wire keys from the backend
+ * `RunCoach.Api.Modules.Training.Constants.WorkoutMetricKeys` — they ARE the
+ * contract, so they mirror the backend verbatim. Labels here are UI-facing
+ * (e.g. `hrAvg → "Avg HR"`) and are intentionally friendlier than the compact
+ * labels the backend `WorkoutMetricKeys.Metadata` uses for LLM one-liners
+ * (`"HR"`); the two maps share the same key set, not the same label strings.
```
leehopper: **💡 [LOW] Doc comment falsely claims the frontend metric map shares the backend's key set**

The file-level comment claims the frontend `WORKOUT_METRIC_META` and the backend `WorkoutMetricKeys.Metadata` "share the same key set." They do not: the frontend map defines 12 keys, while the backend `Metadata` defines 15 — it additionally reserves `verticalOscillation`, `groundContactTime`, and `strideLength`. The backend is a strict superset, so a reader relying on this comment to assume enforced parity would be misled.

---
*Confidence: 75% | deep-review*

```suggestion
 * (`"HR"`). The frontend map is a subset of the backend `Metadata` key set (which also reserves `verticalOscillation`, `groundContactTime`, `strideLength`); the two use different label strings.
```

frontend/src/app/api/generated/index.ts
```diff
@@ -52,6 +53,23 @@ export type OnboardingProgressDto = z.infer<typeof onboardingProgressSchema>
 export const suggestedInputTypeSchema = PostApiV1OnboardingTurnsResponse.shape.suggestedInputType
 export type SuggestedInputType = z.infer<typeof suggestedInputTypeSchema>
 
+// Workout-log schemas (slice-2b). The create-request body is the drift seam
+// the `/log` form derives from via `.pick().extend()` (DEC-075): the form keeps
+// `occurredOn`'s ISO-date format and `completionStatus`'s `0|1|2` enum honest
+// against the backend contract, and a backend rename surfaces here as a TS
+// error during `npm run codegen:check`.
```
leehopper: **💡 [LOW] Comment attributes the drift signal to the wrong command (codegen:check runs no tsc)**

The comment says a backend rename "surfaces here as a TS error during `npm run codegen:check`." But `codegen:check` runs `npm run codegen && git diff --exit-code` (no `tsc`): a backend rename is caught there as a generated-output **drift (git-diff) failure**, while the actual TypeScript error from the broken `createWorkoutLogRequestSchema.shape.completionStatus` access surfaces under `npm run build` (`tsc -b`). The comment conflates the drift gate with the type-check gate. (The older onboarding comment above has the same phrasing — out of scope here, but worth a follow-up sweep.)

---
*Confidence: 75% | deep-review*

```suggestion
// against the backend contract. A backend rename is caught by `codegen:check`
// as a generated-output drift (git-diff) failure; the resulting broken `.shape`
// access then surfaces as a TypeScript error under `npm run build`.
```

frontend/src/app/modules/logging/schemas/workout-log-form.schema.ts
```diff
@@ -0,0 +1,174 @@
...
+    const value = values[key]
+    if (value !== undefined) {
+      metrics[key] = value
+    }
+  }
+
+  return {
+    idempotencyKey,
+    occurredOn: values.occurredOn,
+    distanceMeters: (values.distanceKm ?? 0) * 1000,
```
leehopper, line 174: **💡 [LOW] No test pins behavior when a Skipped workout still carries typed distance/duration**

`toCreateWorkoutLogRequest` sends `distanceMeters: (values.distanceKm ?? 0) * 1000`. The form-level `superRefine` lets a Skipped workout *omit* distance/duration but does not *clear* typed values, so the sequence "type 5 in Distance → select Skipped → submit" sends `distanceMeters: 5000` for a skipped run. The fields render unconditionally and are never reset on status change, so this is reachable through the UI. The only Skipped tests use blank fields (asserting the 0-fallback); none pins behavior for a Skipped run carrying a typed non-zero distance, so the intended semantics are unspecified.

**Suggested fix:** Decide the intended semantics — either zero out distance/duration when `completionStatus === Skipped` inside `toCreateWorkoutLogRequest`, or deliberately accept pass-through — and add a test that pins it.

---
*Confidence: 25% | deep-review*

leehopper (commented):
### Deep Review — PR #156 (workout-log `/log` route + logging form)

Frontier-mode multi-agent review (7 dimensions → deterministic verify → Sonnet validate → Opus blind challenge). **4 low-severity findings** posted inline: 1 correctness/observability item and 3 documentation/test-coverage suggestions. Security, cross-file impact, and code simplification came back clean, and the test suite is strong for AI-authored code.

Eliminated by the pipeline (not posted): a hypothetical false save-failed error (unreachable — `toast`/`navigate` don't throw synchronously; render errors hit `AppErrorBoundary`), a float-precision conversion artifact (negligible), and three latent/marginal items below the confidence threshold.

Generated by deep-review | Reviewed up to: b4e83752e12777510689a253c1ddef0517bb7065

frontend/src/app/modules/logging/pages/log.page.tsx
leehopper: Applied (5844425). Bound the error and forwarded it to `reportClientError({ kind: 'unhandled-rejection', error })` before the user-facing alert — the awaited `.unwrap()` rejection is a *handled* rejection, so it was invisible to both `useGlobalErrorReporter` and `AppErrorBoundary`; this restores a diagnostic trail. Added a test asserting the reporter is called on failure. Left the sibling `regenerate-plan-dialog` untouched (out of this PR's scope — different feature), worth the same treatment in a follow-up.

frontend/src/app/modules/logging/metric-meta.ts
leehopper: Fixed (5844425). Confirmed the backend `WorkoutMetricKeys.Metadata` defines 15 keys vs the frontend map's 12 (the backend additionally reserves `verticalOscillation`, `groundContactTime`, `strideLength`). Reworded the doc to state the frontend map is a *subset* of the backend key set, not an equal set.

frontend/src/app/api/generated/index.ts
leehopper: Fixed (5844425). Verified `codegen:check` runs `npm run codegen && git diff --exit-code` (no tsc). Reworded the workout-log comment to say a backend rename is caught by `codegen:check` as a generated-output drift (git-diff) failure, with the broken `.shape` access surfacing as a TS error under `npm run build`. Also corrected the identical misstatement in the onboarding comment above (same file, same conflation) so the follow-up sweep isn't needed.

frontend/src/app/modules/logging/schemas/workout-log-form.schema.ts
leehopper: Addressed (5844425). Decided semantics: distance/duration are deliberate **pass-through**, not status-gated. The inputs stay rendered on every status (never hidden or reset when Skipped is selected), so a typed value remains visible to the user — zeroing it would be a WYSIWYG violation in the other direction. Documented this in the `toCreateWorkoutLogRequest` doc and added a test pinning that a Skipped workout with a typed `distanceKm: 5` ships `distanceMeters: 5000`.

PR #161, opened 2026-06-07
feat(slice-3): deterministic safety gate keyword classifier (PR1)

leehopper:
## Slice 3 — PR1: Deterministic SafetyGate (Unit 3)

First of 7 stacked PRs for Slice 3 (Adaptation Loop). Ships the deterministic safety classifier that runs on a logged workout's free-text **before** any LLM call (DEC-019 / DEC-030 / DEC-079, high-risk subset). **No LLM, no Marten** — a root PR. PR2/PR4/PR5 reference `SafetyTier`, PR5 wires the gate into the orchestration handler, and PR7 consumes the `RecentLogSanitizer` seam.

### What's here
- **`SafetyTier` / `ReferralCategory` enums** + **`SafetyClassification`** result — a sealed record with `Green()` / `Amber()` / `Red()` factories so a Green tier can never carry a referral category (mirrors `OnboardingTurnOutputValidationResult`).
- **`SafetyKeywordCatalog`** — a versioned (`PolicyVersion` v1.1.0), in-code immutable rule catalog mirroring `PatternCatalog`: word-boundary, ReDoS-guarded (50 ms), case-insensitive regexes mapping crisis / emergency-referral / injury / RED-S phrasings to `{tier, category}`, ordered by escalation precedence. Proximity rules bridge with `[\s\S]` so signals survive multi-line notes.
- **`SafetyGate.Classify`** — precedence-ordered, highest-tier-wins, pure/deterministic (Training layer, **zero Coaching dependency**); scans the note plus the free-text `weather`/`terrain` metric values only (numeric metrics are not prose).
- **`CrisisResponseContent`** — the versioned scripted (non-LLM) Red-crisis turn with the exact `988 Suicide & Crisis Lifeline` and `Crisis Text Line: text 741741` resources (empathetic, no probing); kept out of the sanitizer so the literal `&` is preserved.
- **`RecentLogSanitizer`** — routes `LoggedWorkoutDetail` notes + free-text metric values through the DEC-059 `IPromptSanitizer` (`TrainingHistoryWorkoutNote` section). This is the sanitizer-coverage seam; the `WorkoutLog → LoggedWorkoutDetail` entity mapper that calls it lands in PR7.
- **DI:** `ISafetyGate` + `IRecentLogSanitizer` registered as singletons.

### Design notes
- **Deterministic-vs-LLM split (DEC-012):** classification is keyword/threshold, never LLM self-policing. The gate lives in the pure `Modules/Training/Safety/` layer with no dependency on the Coaching/LLM layer.
- **Recall over precision (DEC-079 audience steer):** under-reaction (missing a real crisis/emergency) is the hard-failure mode for the MVP-0 self+family audience; over-reaction is the lesser evil. Following `PatternCatalog`'s false-positive discipline, bare "want to die" is omitted (running hyperbole), self-injury phrasings require an intent/deliberate frame ("tripped and hurt myself" stays Green), and "cutting" requires "myself" so it never fires on "cutting back my mileage".
- A multi-agent **adversarial review** (11 verified findings) hardened crisis / cardiac / hip-femoral / pregnancy / amenorrhea coverage and fixed a multi-line proximity bug before this PR opened.

### Tests
96 unit tests (strict TDD red→green) covering every `safety-gate.feature` scenario + the broadened phrasings + false-positive guards + multi-line proximity. Full backend suite green (**1356/1356**); Release build clean; no OpenAPI drift.

### Follow-ups found (deferred to Unit 6 eval calibration)
- **Negation handling** — "no chest pain" still escalates to Red today. Deferred deliberately: a naive preceding-negator guard would risk *suppressing* a real signal ("not going to lie, I want to kill myself"), so it needs calibrated design against the 5 `TestProfiles`, not a rushed regex. Documented in the catalog `<remarks>`.
- **Fine-grained false-positive precision** (comma-clause-crossing proximity; "water broke" obstetric vocabulary) — fold into the Unit 6 eval-calibration pass where asymmetric scoring (under-reaction = hard fail; over-reaction = low score) tunes thresholds against real profiles.

## Summary by CodeRabbit

* **New Features**
  * Automated safety detection for workout logs that classifies signals into Green/Amber/Red and routes responses accordingly
  * Sanitization of user-provided workout notes and selected free-text metrics to remove harmful characters and contain prompt-like content
  * Added scripted crisis-response content including exact emergency resource strings

* **Tests**
  * Comprehensive unit tests covering sanitization, safety classification rules, keyword catalog, and crisis-response content

### DIFF
--- a/backend/src/RunCoach.Api/Infrastructure/ServiceCollectionExtensions.cs
+++ b/backend/src/RunCoach.Api/Infrastructure/ServiceCollectionExtensions.cs
@@ -4,6 +4,7 @@ using RunCoach.Api.Modules.Coaching.Prompts;
 using RunCoach.Api.Modules.Coaching.Sanitization;
 using RunCoach.Api.Modules.Training.Computations;
 using RunCoach.Api.Modules.Training.Plan;
+using RunCoach.Api.Modules.Training.Safety;
 using RunCoach.Api.Modules.Training.Workouts;
 
 namespace RunCoach.Api.Infrastructure;
@@ -37,6 +38,11 @@ public static class ServiceCollectionExtensions
         services.AddSingleton<IPaceZoneCalculator, PaceZoneCalculator>();
         services.AddSingleton<IHeartRateZoneCalculator, HeartRateZoneCalculator>();
 
+        // Training module — deterministic safety gate (Slice 3 Unit 3 / DEC-079).
+        // Stateless keyword classifier; the keyword catalog is compiled once at
+        // type-load. Consumed by the Slice 3 adaptation orchestration (PR5).
+        services.AddSingleton<ISafetyGate, SafetyGate>();
+
         // Training module — workout-log persistence (scoped, shares the request DbContext).
         services.AddScoped<IWorkoutLogRepository, WorkoutLogRepository>();
         services.AddScoped<IWorkoutLogService, WorkoutLogService>();
@@ -64,6 +70,12 @@ public static class ServiceCollectionExtensions
         // Stateless layered sanitizer — singleton-safe; the pattern catalog
         // is precompiled once at type-load.
         services.AddSingleton<IPromptSanitizer, LayeredPromptSanitizer>();
+
+        // Recent-log free-text sanitizer coverage (Slice 3 Unit 3). Routes
+        // LoggedWorkoutDetail notes + free-text metric values through the DEC-059
+        // sanitizer before the recent-log prompt path; consumed by the Slice 3/4
+        // WorkoutLog → LoggedWorkoutDetail mapper (PR7).
+        services.AddSingleton<IRecentLogSanitizer, RecentLogSanitizer>();
 
         // Idempotency primitive (DEC-060) — scoped so Wolverine handlers and
         // the store share the same `IDocumentSession` instance per request.
--- a/backend/src/RunCoach.Api/Infrastructure/ServiceCollectionExtensions.cs
+++ b/backend/src/RunCoach.Api/Infrastructure/ServiceCollectionExtensions.cs
@@ -40,7 +40,7 @@ public static class ServiceCollectionExtensions
 
         // Training module — deterministic safety gate (Slice 3 Unit 3 / DEC-079).
         // Stateless keyword classifier; the keyword catalog is compiled once at
-        // type-load. Consumed by the Slice 3 adaptation orchestration (PR5).
+        // type-load. Consumed by the Slice 3 adaptation orchestration handler.
         services.AddSingleton<ISafetyGate, SafetyGate>();
 
         // Training module — workout-log persistence (scoped, shares the request DbContext).
@@ -74,7 +74,7 @@ public static class ServiceCollectionExtensions
         // Recent-log free-text sanitizer coverage (Slice 3 Unit 3). Routes
         // LoggedWorkoutDetail notes + free-text metric values through the DEC-059
         // sanitizer before the recent-log prompt path; consumed by the Slice 3/4
-        // WorkoutLog → LoggedWorkoutDetail mapper (PR7).
+        // WorkoutLog → LoggedWorkoutDetail mapper.
         services.AddSingleton<IRecentLogSanitizer, RecentLogSanitizer>();
 
         // Idempotency primitive (DEC-060) — scoped so Wolverine handlers and
--- a/backend/src/RunCoach.Api/Modules/Coaching/Sanitization/RecentLogSanitizer.cs
+++ b/backend/src/RunCoach.Api/Modules/Coaching/Sanitization/RecentLogSanitizer.cs
@@ -1,0 +1,64 @@
+using RunCoach.Api.Modules.Training.Constants;
+using RunCoach.Api.Modules.Training.Models;
+
+namespace RunCoach.Api.Modules.Coaching.Sanitization;
+
+/// <summary>
+/// Default <see cref="IRecentLogSanitizer"/> — sanitizes the note and free-text
+/// metric values of a <see cref="LoggedWorkoutDetail"/> via the DEC-059
+/// <see cref="IPromptSanitizer"/> under the
+/// <see cref="PromptSection.TrainingHistoryWorkoutNote"/> policy. Stateless;
+/// registered as a singleton.
+/// </summary>
+public sealed class RecentLogSanitizer(IPromptSanitizer sanitizer) : IRecentLogSanitizer
+{
+    private static readonly string[] FreeTextMetricKeys =
+        [WorkoutMetricKeys.Weather, WorkoutMetricKeys.Terrain];
+
+    private readonly IPromptSanitizer _sanitizer = sanitizer;
+
+    /// <inheritdoc />
+    public async ValueTask<LoggedWorkoutDetail> SanitizeAsync(
+        LoggedWorkoutDetail detail,
+        CancellationToken ct = default)
+    {
+        ArgumentNullException.ThrowIfNull(detail);
+
+        var notes = detail.Notes;
+        if (!string.IsNullOrEmpty(notes))
+        {
+            var sanitizedNote = await _sanitizer
+                .SanitizeAsync(notes, PromptSection.TrainingHistoryWorkoutNote, ct)
+                .ConfigureAwait(false);
+            notes = sanitizedNote.Sanitized;
+        }
+
+        var metrics = await SanitizeFreeTextMetricsAsync(detail.Metrics, ct).ConfigureAwait(false);
+
+        return detail with { Notes = notes, Metrics = metrics };
+    }
+
+    private async ValueTask<IReadOnlyDictionary<string, string>> SanitizeFreeTextMetricsAsync(
+        IReadOnlyDictionary<string, string> metrics,
+        CancellationToken ct)
+    {
+        if (!metrics.ContainsKey(WorkoutMetricKeys.Weather) && !metrics.ContainsKey(WorkoutMetricKeys.Terrain))
+        {
+            return metrics;
+        }
+
+        var sanitized = new Dictionary<string, string>(metrics, StringComparer.Ordinal);
+        foreach (var key in FreeTextMetricKeys)
+        {
+            if (sanitized.TryGetValue(key, out var value) && !string.IsNullOrEmpty(value))
+            {
+                var result = await _sanitizer
+                    .SanitizeAsync(value, PromptSection.TrainingHistoryWorkoutNote, ct)
+                    .ConfigureAwait(false);
+                sanitized[key] = result.Sanitized;
+            }
+        }
+
+        return sanitized;
+    }
+}
--- a/backend/src/RunCoach.Api/Modules/Coaching/Sanitization/RecentLogSanitizer.cs
+++ b/backend/src/RunCoach.Api/Modules/Coaching/Sanitization/RecentLogSanitizer.cs
@@ -42,7 +42,7 @@ public sealed class RecentLogSanitizer(IPromptSanitizer sanitizer) : IRecentLogS
         IReadOnlyDictionary<string, string> metrics,
         CancellationToken ct)
     {
-        if (!metrics.ContainsKey(WorkoutMetricKeys.Weather) && !metrics.ContainsKey(WorkoutMetricKeys.Terrain))
+        if (!FreeTextMetricKeys.Any(metrics.ContainsKey))
         {
             return metrics;
         }
--- a/backend/src/RunCoach.Api/Modules/Training/Safety/CrisisResponseContent.cs
+++ b/backend/src/RunCoach.Api/Modules/Training/Safety/CrisisResponseContent.cs
@@ -12,7 +12,7 @@ namespace RunCoach.Api.Modules.Training.Safety;
 /// verbatim by tests.
 /// </summary>
 /// <remarks>
-/// Encodes the persona crisis protocol (coaching-persona.md): acknowledge with
+/// Encodes the persona crisis protocol: acknowledge with
 /// empathy, provide the two crisis resources, normalize help-seeking, never
 /// probe for plans or methods, and do not continue the crisis topic. The full
 /// pre-public-release safety scaffolding (PAR-Q+, expanded taxonomy) is deferred.
--- a/backend/src/RunCoach.Api/Modules/Training/Safety/CrisisResponseContent.cs
+++ b/backend/src/RunCoach.Api/Modules/Training/Safety/CrisisResponseContent.cs
@@ -1,0 +1,42 @@
+namespace RunCoach.Api.Modules.Training.Safety;
+
+/// <summary>
+/// The versioned, scripted (non-LLM) coach response surfaced when
+/// <see cref="ISafetyGate"/> classifies a log as <see cref="SafetyTier.Red"/>
+/// with <see cref="ReferralCategory.Crisis"/> (Slice 3 Unit 3 / DEC-079, US
+/// locale, MVP-0). Deterministic system-authored copy — it must NOT pass
+/// through the prompt sanitizer (the literal ampersand in
+/// "988 Suicide &amp; Crisis Lifeline" would be escaped). The exact resource
+/// strings <c>988 Suicide &amp; Crisis Lifeline</c> and
+/// <c>Crisis Text Line: text 741741</c> are contractually required and asserted
+/// verbatim by tests.
+/// </summary>
+/// <remarks>
+/// Encodes the persona crisis protocol (coaching-persona.md): acknowledge with
+/// empathy, provide the two crisis resources, normalize help-seeking, never
+/// probe for plans or methods, and do not continue the crisis topic. The full
+/// pre-public-release safety scaffolding (PAR-Q+, expanded taxonomy) is deferred.
+/// </remarks>
+public static class CrisisResponseContent
+{
+    /// <summary>
+    /// Content version stamped onto the emitted safety turn so audit replay can
+    /// pin the active scripted copy. Bump on any wording change.
+    /// </summary>
+    public const string ContentVersion = "v1.0.0";
+
+    /// <summary>
+    /// The scripted Red-crisis coach turn. Contains the exact required strings
+    /// <c>988 Suicide &amp; Crisis Lifeline</c> and <c>Crisis Text Line: text 741741</c>.
+    /// </summary>
+    public const string CrisisResponse =
+        "It sounds like you're carrying something really heavy right now, and I'm "
+        + "glad you put it into words. This matters more than any run, and you "
+        + "deserve real support from people trained to help. Please reach out right "
+        + "now: the 988 Suicide & Crisis Lifeline (call or text 988), or the "
+        + "Crisis Text Line: text 741741. Reaching out is a strong, healthy thing "
+        + "to do, and you don't have to do it alone. I'm your running coach, so this "
+        + "is outside what I can help with — but the people at those numbers are "
+        + "there for you any time, day or night. I'll be right here for your "
+        + "training whenever you're ready to come back to it.";
+}
--- a/backend/src/RunCoach.Api/Modules/Training/Safety/SafetyClassification.cs
+++ b/backend/src/RunCoach.Api/Modules/Training/Safety/SafetyClassification.cs
@@ -1,0 +1,62 @@
+namespace RunCoach.Api.Modules.Training.Safety;
+
+/// <summary>
+/// The result of <see cref="ISafetyGate.Classify"/>: a <see cref="SafetyTier"/>
+/// paired with the <see cref="ReferralCategory"/> that drove it. PII-free —
+/// carries only the tier + category, never the matched note text.
+/// </summary>
+/// <remarks>
+/// The constructor is <c>internal</c>; callers construct via <see cref="Green"/>,
+/// <see cref="Amber"/>, or <see cref="Red"/> so contradictory pairings (a
+/// non-Green tier with <see cref="ReferralCategory.None"/>, or a Green tier
+/// carrying a category) cannot be expressed. Mirrors
+/// <c>OnboardingTurnOutputValidationResult</c>.
+/// </remarks>
+public sealed record SafetyClassification
+{
+    /// <summary>
+    /// Constructs a classification directly from its components. Internal so
+    /// production callers route through the factories; the test assembly
+    /// retains access via <c>InternalsVisibleTo</c>.
+    /// </summary>
+    /// <param name="tier">The resolved safety tier.</param>
+    /// <param name="category">The referral category driving the tier.</param>
+    internal SafetyClassification(SafetyTier tier, ReferralCategory category)
+    {
+        Tier = tier;
+        Category = category;
+    }
+
+    /// <summary>Gets the resolved safety tier.</summary>
+    public SafetyTier Tier { get; }
+
+    /// <summary>Gets the referral category driving the tier (<see cref="ReferralCategory.None"/> when Green).</summary>
+    public ReferralCategory Category { get; }
+
+    /// <summary>Returns the no-signal result: <see cref="SafetyTier.Green"/> + <see cref="ReferralCategory.None"/>.</summary>
+    public static SafetyClassification Green() => new(SafetyTier.Green, ReferralCategory.None);
+
+    /// <summary>
+    /// Returns an <see cref="SafetyTier.Amber"/> result for the given referral category.
+    /// </summary>
+    /// <param name="category">The driving category. Must not be <see cref="ReferralCategory.None"/>.</param>
+    public static SafetyClassification Amber(ReferralCategory category) => Create(SafetyTier.Amber, category);
+
+    /// <summary>
+    /// Returns a <see cref="SafetyTier.Red"/> result for the given referral category.
+    /// </summary>
+    /// <param name="category">The driving category. Must not be <see cref="ReferralCategory.None"/>.</param>
+    public static SafetyClassification Red(ReferralCategory category) => Create(SafetyTier.Red, category);
+
+    private static SafetyClassification Create(SafetyTier tier, ReferralCategory category)
+    {
+        if (category == ReferralCategory.None)
+        {
+            throw new ArgumentException(
+                $"{tier} requires a non-None referral category; use Green() for the no-signal case.",
+                nameof(category));
+        }
+
+        return new SafetyClassification(tier, category);
+    }
+}
--- a/backend/src/RunCoach.Api/Modules/Training/Safety/SafetyClassification.cs
+++ b/backend/src/RunCoach.Api/Modules/Training/Safety/SafetyClassification.cs
@@ -57,6 +57,24 @@ public sealed record SafetyClassification
                 nameof(category));
         }
 
+        // Enforce the tier-category pairing the enum docs assert: `Crisis` and
+        // `EmergencyReferral` are always Red, `Injury` and `RedS` always Amber.
+        // A mis-paired catalog rule fails construction rather than emitting a
+        // contradictory classification.
+        var validForTier = tier switch
+        {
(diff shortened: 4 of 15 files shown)


backend/src/RunCoach.Api/Modules/Training/Safety/SafetyKeywordCatalog.cs
```diff
@@ -0,0 +1,181 @@
...
+        Rule("SG-E07", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\b(?:passed\s+out|fainted|blacked\s+out|syncope|(?:going\s+to|about\s+to|gonna|nearly|almost)\s+(?:faint|pass\s+out|black\s+out))\b"),
+        Rule("SG-E08", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\bfemoral\b"),
+        Rule("SG-E09", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\bgroin\b[\s\S]{0,12}?\b(?:pain|hurt\w*|ache\w*)\b"),
+        Rule("SG-E10", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\b(?:pain|hurt\w*|ache\w*)\b[\s\S]{0,12}?\bgroin\b"),
+        Rule("SG-E11", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\bhip\b[\s\S]{0,30}?\b(?:worse|worsening|gets?\s+worse|getting\s+worse|every\s+(?:run|step|stride)|deep\s+ache|stress\s+fracture|can['\u2019]?t\s+(?:bear|put)\s+(?:any\s+)?weight)\b"),
+        Rule("SG-E12", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\b(?:sharp\s+pain|deep\s+ache|stabbing)\b[\s\S]{0,12}?\bhip\b"),
+        Rule("SG-E13", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\bvaginal\s+bleeding\b"),
```
leehopper, line 121: **🟡 [MEDIUM] SG-E13 ("vaginal bleeding") has no test row**

SG-E13 is a standalone Red/EmergencyReferral rule with no pregnancy precondition, but no test exercises it: every bleeding/spotting/contractions row in `SafetyGateTests` also carries a pregnancy marker and therefore matches SG-E14/E15, not E13. Bleeding *without* a stated pregnancy is exactly the case E13 exists to catch (recall-over-precision), and it is entirely unverified.

**Suggested fix:** Add an emergency `[InlineData]` row such as `"heavy vaginal bleeding during the run"` (no pregnancy word) and assert `Red/EmergencyReferral`, isolating E13.

*Confidence: 100% · deep-review*

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyKeywordCatalog.cs
```diff
@@ -0,0 +1,181 @@
...
+    private static ImmutableArray<SafetyRule> CrisisRules() =>
+    [
+        Rule("SG-C01", SafetyTier.Red, ReferralCategory.Crisis, @"\bkill(?:ing)?\s+myself\b"),
+        Rule("SG-C02", SafetyTier.Red, ReferralCategory.Crisis, @"\bend(?:ing)?\s+my\s+(?:own\s+)?life\b"),
+        Rule("SG-C03", SafetyTier.Red, ReferralCategory.Crisis, @"\bend(?:ing)?\s+it\s+all\b"),
+        Rule("SG-C04", SafetyTier.Red, ReferralCategory.Crisis, @"\bself[\s-]?harm(?:ing|ed)?\b"),
+        Rule("SG-C05", SafetyTier.Red, ReferralCategory.Crisis, @"\bsuicidal\b"),
+        Rule("SG-C06", SafetyTier.Red, ReferralCategory.Crisis, @"\bsuicide\b"),
+        Rule("SG-C07", SafetyTier.Red, ReferralCategory.Crisis, @"\b(?:want(?:s|ing|ed)?|going|trying|urge)\s+to\s+(?:hurt|harm)\s+myself\b"),
+        Rule("SG-C08", SafetyTier.Red, ReferralCategory.Crisis, @"\b(?:thinking\s+about\s+(?:hurting|harming)\s+myself|(?:hurt|harm)(?:ing)?\s+myself\s+(?:on\s+purpose|deliberately|intentionally))\b"),
```
leehopper, line 96: **🟡 [MEDIUM] Crisis rules SG-C08 / SG-C09 (intent-qualified self-harm) have no positive test row**

The 12 crisis `[InlineData]` rows cover SG-C01–C07 and C10–C14. SG-C08 ("thinking about hurting myself" / "hurt myself on purpose|deliberately|intentionally") and SG-C09 ("cutting myself" / "cut myself on purpose") have only their *negative* side tested ("tripped and hurt myself", "cut myself on a branch" stay Green). Nothing asserts the deliberate-intent phrasing actually escalates to Red — and these are the two most regex-intricate crisis rules.

**Suggested fix:** Add crisis rows e.g. `"I keep thinking about hurting myself"`, `"I cut myself on purpose after the run"`, `"been cutting myself again"` asserting `Red/Crisis`.

*Confidence: 73% · deep-review*

backend/tests/RunCoach.Api.Tests/Modules/Training/Safety/SafetyKeywordCatalogTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        var firstAppearance = Catalog.Rules.Select(r => r.Category).Distinct().ToList();
+        firstAppearance.Should().Equal(
+            ReferralCategory.Crisis,
+            ReferralCategory.EmergencyReferral,
+            ReferralCategory.Injury,
+            ReferralCategory.RedS);
+    }
+
+    [Fact]
+    public void ForTesting_BacksTheCatalogWithTheGivenRules()
```
leehopper, line 79: **🟡 [MEDIUM] Regex-timeout fail-open path is never exercised by any test**

`SafetyGate.Matches` catches `RegexMatchTimeoutException` and fails **open** (returns non-match → Green) — the single most safety-critical branch in this file. The `ForTesting` factory and internal `SafetyGate(catalog)` ctor exist specifically to test it (the XML-doc even cites "ReDoS timeout handling"), and the codebase already has the poison-regex pattern in `LayeredPromptSanitizerTests`. But the only `ForTesting` test just round-trips a RuleId; the catch is uncovered.

**Suggested fix:** Build a catalog via `ForTesting` with a poison rule (`new Regex("(a+)+$", Compiled, TimeSpan.FromTicks(1))`, Red/Crisis), construct `new SafetyGate(catalog)`, call `Classify` with ~30 a-chars + `!`, and assert it does not throw and returns the intended tier — locking in (or surfacing) the fail-open behavior. See also bug-1.

*Confidence: 70% · deep-review*

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyKeywordCatalog.cs
```diff
@@ -0,0 +1,181 @@
...
+        Rule("SG-R03", SafetyTier.Amber, ReferralCategory.RedS, @"\b(?:haven['\u2019]?t\s+had|hasn['\u2019]?t\s+(?:had|come)|lost|stopped\s+having|no\s+(?:more\s+)?)\s+(?:\w+\s+){0,3}periods?\b"),
+        Rule("SG-R04", SafetyTier.Amber, ReferralCategory.RedS, @"\b(?:my\s+)?periods?\b[\s\S]{0,20}?\b(?:stopped|disappeared|gone|hasn['\u2019]?t\s+come|haven['\u2019]?t\s+come)\b"),
+        Rule("SG-R05", SafetyTier.Amber, ReferralCategory.RedS, @"\bstress\s+fracture\b"),
+        Rule("SG-R06", SafetyTier.Amber, ReferralCategory.RedS, @"\b(?:not\s+eating\s+enough|under[\s-]?eating|not\s+fueling|not\s+eating\s+back|restrict\w*\s+(?:my\s+)?(?:food|eating|calories|intake))\b"),
+        Rule("SG-R07", SafetyTier.Amber, ReferralCategory.RedS, @"\bearn(?:ing|ed)?\s+(?:my|the)\s+(?:food|meal|calories|dinner|carbs)\b"),
+        Rule("SG-R08", SafetyTier.Amber, ReferralCategory.RedS, @"\b(?:run|running|ran)\s+(?:it|them|that)\s+off\b(?![\s\S]{0,8}\b(?:road|trail|track|path|course|back|front|pack)\b)"),
+        Rule("SG-R09", SafetyTier.Amber, ReferralCategory.RedS, @"\bburn(?:ing|ed)?\s+(?:it|them|those|the\s+calories)\s+off\b"),
```
leehopper, line 152: **🟡 [MEDIUM] RED-S rule SG-R09 ("burn it/the calories off") has no positive test row**

SG-R09 matches compensatory-exercise phrasing "burn(ing) it/them/those/the calories off". The only `burn` string in the tests is the benign Green row "burned a ton of calories on the long run" (which correctly does not match R09). Nothing asserts a disordered-pattern positive escalates Amber, while the sibling SG-R08 ("run it off") *is* tested.

**Suggested fix:** Add a RED-S row e.g. `"ate too much so I went out to burn it off"` asserting `Amber/RedS`.

*Confidence: 67% · deep-review*

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Sanitization/RecentLogSanitizerTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        // Act
+        var result = await _sut.SanitizeAsync(detail, TestContext.Current.CancellationToken);
+
+        // Assert — the note is wrapped in the Spotlighting containment delimiter.
+        result.Notes.Should().Contain("WORKOUT_NOTE");
+    }
+
+    [Fact]
+    public async Task SanitizeAsync_SanitizesFreeTextMetricValues()
```
leehopper, line 50: **🟡 [MEDIUM] Free-text metric path untested for injection containment (only zero-width strip)**

The Notes path has two tests — zero-width strip AND injection containment (asserts the `WORKOUT_NOTE` delimiter wraps an "ignore all previous instructions" payload). The free-text metric path (`SanitizeFreeTextMetricsAsync`) has only the zero-width test; no test asserts an injection payload in a weather/terrain value gets the same containment, even though `safety-gate.feature` explicitly names metric values as an injection surface. (Production code does apply it — coverage gap, not behavior gap.)

**Suggested fix:** Add a test setting `metrics[Weather]` to an injection payload and assert the result contains `WORKOUT_NOTE`, mirroring the note injection test.

*Confidence: 57% · deep-review*

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyKeywordCatalog.cs
```diff
@@ -0,0 +1,181 @@
...
+        Rule("SG-E03", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\b(?:tight\w*|pressure|squeez\w*|heaviness)\b[\s\S]{0,20}?\bchest\b"),
+        Rule("SG-E04", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\birregular\s+(?:heart\s*beat|pulse|heart\s+rhythm)\b"),
+        Rule("SG-E05", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\b(?:palpitations?|skip(?:ping|ped)\s+(?:a\s+)?beats?|heart\s+(?:\w+\s+){0,2}(?:skip\w*|flutter\w*))\b"),
+        Rule("SG-E06", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\bpain\b[\s\S]{0,30}?\b(?:left\s+arm|down\s+my\s+arm|jaw|radiat\w*)\b"),
+        Rule("SG-E07", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\b(?:passed\s+out|fainted|blacked\s+out|syncope|(?:going\s+to|about\s+to|gonna|nearly|almost)\s+(?:faint|pass\s+out|black\s+out))\b"),
+        Rule("SG-E08", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\bfemoral\b"),
+        Rule("SG-E09", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\bgroin\b[\s\S]{0,12}?\b(?:pain|hurt\w*|ache\w*)\b"),
+        Rule("SG-E10", SafetyTier.Red, ReferralCategory.EmergencyReferral, @"\b(?:pain|hurt\w*|ache\w*)\b[\s\S]{0,12}?\bgroin\b"),
```
leehopper, line 118: **🟡 [MEDIUM] Reverse-order proximity rules (E10/E12/E15/I04/I06/R12) are masked, never isolated**

Several rules exist in forward/reverse pairs to catch both word orders. `Classify` returns on first match in catalog order, every forward rule precedes its reverse sibling, and every test input uses the forward order — so the reverse rules (E10 pain→groin, E12 sharp-pain→hip, E15 bleed→pregnant, I04 worse→pain, I06 pain→curtailment, R12 rest-day→distress) never decide a result. The category assertion still passes because siblings share Tier+Category, masking a broken reverse regex.

**Suggested fix:** Add one isolating row per reverse rule with the opposite word order, e.g. E15 `"bleeding badly, I am 12 weeks pregnant"`; I04 `"it is getting worse, real pain now"`; R12 `"took a rest day and felt so guilty afterwards"`.

*Confidence: 55% · deep-review*

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyClassification.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <param name="category">The driving category. Must not be <see cref="ReferralCategory.None"/>.</param>
+    public static SafetyClassification Amber(ReferralCategory category) => Create(SafetyTier.Amber, category);
+
+    /// <summary>
+    /// Returns a <see cref="SafetyTier.Red"/> result for the given referral category.
+    /// </summary>
+    /// <param name="category">The driving category. Must not be <see cref="ReferralCategory.None"/>.</param>
+    public static SafetyClassification Red(ReferralCategory category) => Create(SafetyTier.Red, category);
+
+    private static SafetyClassification Create(SafetyTier tier, ReferralCategory category)
```
leehopper, line 51: **🟡 [MEDIUM] Factories enforce only the None-check, not the documented tier↔category pairing**

The internal-ctor + Green/Amber/Red factories correctly make the None-pairing contradictions unrepresentable, but the finer mapping the enum XML-docs assert (Crisis/EmergencyReferral ⇒ Red only; Injury/RedS ⇒ Amber only) is **not** enforced: `Red(ReferralCategory.Injury)` and `Amber(ReferralCategory.Crisis)` both construct. `SafetyGate` reads `rule.Tier` and `rule.Category` independently and never cross-checks, so a mis-authored catalog rule would silently emit a contradictory classification. Latent today (the production catalog is internally consistent).

**Suggested fix:** In `Create()`, switch on tier and reject categories outside the legal set (Red⇒{Crisis,EmergencyReferral}, Amber⇒{Injury,RedS}), turning the doc-only pairing into an enforced construction invariant.

*Confidence: 20% · deep-review*

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyGate.cs
```diff
@@ -0,0 +1,82 @@
...
+        if (string.IsNullOrWhiteSpace(text))
+        {
+            return false;
+        }
+
+        try
+        {
+            return rule.Matcher.IsMatch(text);
+        }
+        catch (RegexMatchTimeoutException)
```
leehopper, line 74: **💡 [LOW] Regex timeout fails OPEN (→ Green) in a recall-prioritized safety gate**

On `RegexMatchTimeoutException`, `Matches` returns false; since `Classify` returns `Green()` when nothing matched, a note that times out is silently classified safe even if it contains a real crisis/emergency signal — a fail-**open** in the unsafe direction, contradicting the DEC-079 / catalog stance that under-reaction is the hard-failure mode. **Not reachable today** (`Classify` has no production caller until PR5), and the bounded quantifiers make an actual 50 ms timeout unlikely — hence low — but worth settling before PR5 wires the gate in.

**Suggested fix:** Fail closed on timeout — record the timed-out ruleId and return a non-Green / needs-review classification, or cap input length before scanning so the timeout path can never drop a real signal.

*Confidence: 70% · deep-review*

backend/src/RunCoach.Api/Modules/Coaching/Sanitization/RecentLogSanitizer.cs
```diff
@@ -0,0 +1,64 @@
...
+        var metrics = await SanitizeFreeTextMetricsAsync(detail.Metrics, ct).ConfigureAwait(false);
+
+        return detail with { Notes = notes, Metrics = metrics };
+    }
+
+    private async ValueTask<IReadOnlyDictionary<string, string>> SanitizeFreeTextMetricsAsync(
+        IReadOnlyDictionary<string, string> metrics,
+        CancellationToken ct)
+    {
+        if (!metrics.ContainsKey(WorkoutMetricKeys.Weather) && !metrics.ContainsKey(WorkoutMetricKeys.Terrain))
```
leehopper: **💡 [LOW] Fast-path guard re-lists the free-text keys instead of reusing FreeTextMetricKeys**

`FreeTextMetricKeys = [Weather, Terrain]` is the single source of truth for the sanitization loop, but the early-return guard re-spells the same set inline (`!metrics.ContainsKey(Weather) && !metrics.ContainsKey(Terrain)`). The two must be kept in sync by hand: a future third free-text key added only to the array would let the guard wrongly take the fast path and skip sanitizing it.

**Suggested fix:** Drive the guard off the array — `if (!FreeTextMetricKeys.Any(metrics.ContainsKey)) { return metrics; }` (add `using System.Linq;` if needed) — preserving the dict-copy-avoidance fast path while making the array the single authority.

*Confidence: 63% · deep-review*

backend/src/RunCoach.Api/Modules/Training/Safety/CrisisResponseContent.cs
```diff
@@ -0,0 +1,42 @@
...
+/// with <see cref="ReferralCategory.Crisis"/> (Slice 3 Unit 3 / DEC-079, US
+/// locale, MVP-0). Deterministic system-authored copy — it must NOT pass
+/// through the prompt sanitizer (the literal ampersand in
+/// "988 Suicide &amp; Crisis Lifeline" would be escaped). The exact resource
+/// strings <c>988 Suicide &amp; Crisis Lifeline</c> and
+/// <c>Crisis Text Line: text 741741</c> are contractually required and asserted
+/// verbatim by tests.
+/// </summary>
+/// <remarks>
+/// Encodes the persona crisis protocol (coaching-persona.md): acknowledge with
```
leehopper: **💡 [LOW] Doc-comment references external doc path coaching-persona.md**

REVIEW.md requires comments to be self-contained and forbids external doc-path references (the `docs/planning/...` shape). The `<remarks>` embeds `(coaching-persona.md)`; the surrounding prose already summarizes the protocol, so the path is not load-bearing.

**Suggested fix:** Remove the path reference (suggestion below). If a durable anchor is wanted, cite a DEC instead of a file path.

*Confidence: 60% · deep-review*

```suggestion
/// Encodes the persona crisis protocol: acknowledge with
```

leehopper (commented):
### Deep Review — Slice 3 deterministic safety gate

Frontier-mode multi-agent review (7 dimensions, blind-challenge verified). The classifier itself is solid: every test row verified to match a catalog rule, **no broken regexes and no ReDoS**, **zero cross-file breakage**, and the Training→Coaching architecture boundary holds.

**11 findings, all low/medium** — no critical or high blocking issues. Posting all 11 as inline comments per request. Themes:

- **Recall-critical safety rules ship without positive test coverage** — SG-E13 `vaginal bleeding` (no row), SG-C08/C09 intent-qualified self-harm, SG-R09 `burn it off`. Given the stated recall-over-precision priority these are the ones worth closing.
- **The regex-timeout path fails *open* (→ Green) and is untested** — latent today (no production caller until PR5 wires `Classify` in), but worth settling before that wiring lands.
- **tier↔category pairing is documented but not enforced** in the SafetyClassification factories.
- Minor convention/simplification nits.

Two security findings (fail-open security framing, per-rule-timeout DoS) were eliminated at the confidence threshold as practically-constrained (linear-slow, authenticated endpoint, input-size-cap already a tracked backlog item).

---
Generated by deep-review | Reviewed up to: 6002ec1889dd2ecfda8fe8480ff742c3ef9e29bb

backend/src/RunCoach.Api/Infrastructure/ServiceCollectionExtensions.cs
```diff
@@ -37,6 +38,11 @@ public static IServiceCollection AddApplicationModules(
         services.AddSingleton<IPaceZoneCalculator, PaceZoneCalculator>();
         services.AddSingleton<IHeartRateZoneCalculator, HeartRateZoneCalculator>();
 
+        // Training module — deterministic safety gate (Slice 3 Unit 3 / DEC-079).
+        // Stateless keyword classifier; the keyword catalog is compiled once at
+        // type-load. Consumed by the Slice 3 adaptation orchestration (PR5).
```
leehopper: **💡 [LOW] DI comments carry (PR5)/(PR7) planning-phase forward references**

REVIEW.md forbids planning-phase forward references and limits the allowlist to Spec/DEC/R-NNN symbolic refs. Line 43 (`...orchestration (PR5)`) and line 77 (`...mapper (PR7)`) point at future work items by bare planning label. The lines do carry valid DEC anchors, so this is borderline (contested) — but the bare `(PR5)`/`(PR7)` tokens are not on the allowlist.

**Suggested fix:** Drop the bare `(PR5)`/`(PR7)` pointers or re-anchor to a durable DEC reference; keep the load-bearing why-deferred prose.

*Confidence: 57% · deep-review*

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyKeywordCatalog.cs
leehopper: Addressed in 28b5d13. Added `"heavy vaginal bleeding during the run"` (no pregnancy marker) to the emergency theory in `SafetyGateTests`. With no marker present, `SG-E14`/`SG-E15` cannot fire, so the row isolates `SG-E13`.

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyKeywordCatalog.cs
leehopper: Addressed in 28b5d13. Added positive crisis rows: `"I keep thinking about hurting myself"` (`SG-C08`), and `"I cut myself on purpose after the run"` + `"been cutting myself again"` (`SG-C09`). Each traces to the intended rule as first-match and asserts Red/Crisis.

backend/tests/RunCoach.Api.Tests/Modules/Training/Safety/SafetyKeywordCatalogTests.cs
leehopper: Addressed in 28b5d13. Added `Classify_FailsClosed_WhenAGateRuleTimesOut` — it builds a `ForTesting` catalog with a poison `(a+)+$` rule (`TimeSpan.FromTicks(1)`, mirroring `LayeredPromptSanitizerTests`), constructs the internal `SafetyGate(catalog)`, and asserts the timeout escalates to the rule tier. Placed in `SafetyGateTests` since it asserts `Classify` behavior. Paired with the fail-closed fix (see the `SafetyGate.cs:74` thread).

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyKeywordCatalog.cs
leehopper: Addressed in 28b5d13. Added `"ate too much so I went out to burn it off"` -> Amber/RedS. `SG-R08` (`run it off`) does not fire on the `burn` verb, so the row isolates `SG-R09`.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Sanitization/RecentLogSanitizerTests.cs
leehopper: Addressed in 28b5d13. Added `SanitizeAsync_ContainsInjectionInFreeTextMetricWithinTheWorkoutNoteDelimiter`, asserting an injection payload in `metrics[Weather]` is wrapped in the `WORKOUT_NOTE` containment delimiter, mirroring the note injection test.

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyKeywordCatalog.cs
leehopper: Addressed in 28b5d13. Added isolating rows for the reverse rules: `SG-E10` (`deep pain in the groin...`), `SG-E12` (`...stabbing in my hip`), `SG-E15` (`bleeding badly, I am 12 weeks pregnant`), `SG-I04` (`getting worse, real pain now`), `SG-I06` (`the pain got so bad I had to stop`), `SG-R12` (`took a rest day and felt so guilty afterwards`). For each, the forward sibling provably does not match, so a broken reverse regex now fails the test rather than being masked.

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyClassification.cs
leehopper: Addressed in 28b5d13. `SafetyClassification.Create` now switches on tier and rejects out-of-set categories (Red => `Crisis`/`EmergencyReferral`, Amber => `Injury`/`RedS`), turning the doc-only pairing into an enforced construction invariant. Added throw-tests in `SafetyClassificationTests`.

backend/src/RunCoach.Api/Modules/Training/Safety/SafetyGate.cs
leehopper: Addressed in 28b5d13. The `RegexMatchTimeoutException` path now fails CLOSED: it returns a match so the gate escalates to the timed-out rule tier rather than silently returning Green. This matches the DEC-079 recall-over-precision stance (under-reaction is the hard-failure mode). Because `Classify` scans in strict precedence order, a timeout can only escalate, never downgrade. Locked in by the new poison-regex test.

backend/src/RunCoach.Api/Modules/Coaching/Sanitization/RecentLogSanitizer.cs
leehopper: Addressed in 28b5d13. The guard now reads `!FreeTextMetricKeys.Any(metrics.ContainsKey)`, so the array is the single source of truth and a future third free-text key is picked up automatically (no hand-sync).

backend/src/RunCoach.Api/Modules/Training/Safety/CrisisResponseContent.cs
leehopper: Addressed in 28b5d13. Removed the `(coaching-persona.md)` path; the surrounding prose already summarizes the protocol, so nothing load-bearing was lost.

backend/src/RunCoach.Api/Infrastructure/ServiceCollectionExtensions.cs
leehopper: Addressed in 28b5d13. Dropped the bare `(PR5)`/`(PR7)` tokens on both lines, keeping the load-bearing consumer/why prose and the DEC-079 / DEC-059 anchors.

leehopper:
@coderabbitai full review

PR #163, opened 2026-06-08
feat(slice-3): adaptation events + projection + conversation read-model (PR3)

leehopper:
Slice 3 (Adaptation Loop) **PR3 of 7 — Unit 2**: the Marten event/projection surface for a logged-workout adaptation plus the read-only conversation read-model and its read endpoint. Roots the merge train alongside PR1 (SafetyGate, #161); ships unreferenced contracts that PR5 wires into production.

Spec: `docs/specs/17-spec-slice-3-adaptation/` (Unit 2 + `pr-strategy.md` + `adaptation-events-and-projection.feature`, gitignored/working-tree-only). Decisions: DEC-079, DEC-058/060/067/012/037.

## What's in it

**Events (Plan stream):**
- `PlanAdaptedFromLog` `{ TriggeringWorkoutLogId, AdaptationKind, EscalationLevel, SafetyTier, Rationale, Diff }` and `SafetySignalRaised` `{ TriggeringWorkoutLogId, SafetyTier, ReferralCategory, Content }`, appended to the existing per-user Plan stream (never `StartStream`), registered with the `_v1` schema-version suffix (DEC-067).
- `EscalationLevel` (DEC-012 0-indexed) + `AdaptationKind` enums, and a **structured** `PlanAdaptationDiff` (`WorkoutChange` / `WeeklyTargetChange`) so the panel renders before/after from structured data, not parsed prose (resolves the Unit 2 open question).

**Projection:**
- `PlanProjection.Apply(PlanAdaptedFromLog)` — swaps current-week micro workouts + revises upcoming meso `WeeklyTargetKm`; **never synthesizes** micro detail for weeks that have none today (spec non-goal); 1-based-week guard fails the transaction loudly.
- Net-new `ConversationLogView` read-model + `ConversationProjection` — an **inline `SingleStreamProjection` over the same Plan stream**, keyed by `PlanId`, emitting one turn per event keyed by the Marten event id for replay idempotency. Distinct from the legacy two-string `Coaching/Models/ConversationTurn` DTO.

**Read endpoint + codegen:**
- `GET /api/v1/conversation/turns` — read-only (no antiforgery), user-scoped via a tenanted Marten session, newest-first, 200-with-empty when there's no active plan or no turns yet (PR5 wires production appends).
- Regenerated `backend/openapi/swagger.json` (+ the path + 10 schemas) and the frontend RTK + Orval-zod output; new DTO/enum/diff types re-exported type-only from the generated barrel so PR7 has them.

## Key design decision (worth a look)

The conversation read-model needs *one logical turn per event*. Rather than introduce a `MultiStreamProjection` fan-out (zero exemplars in this repo → Research-Protocol territory), this uses an **accumulating `SingleStreamProjection<ConversationLogView, Guid>` keyed by the Plan stream id** — the precedented `PlanProjection` pattern. One per-plan conversation log; a regenerated plan starts a fresh log (correct panel semantics). The read endpoint mirrors `PlanRenderingController`'s exact EF-resolve-active-plan → tenanted-Marten-load recipe. Enums are int-encoded (no `JsonStringEnumConverter`), consistent with the deliberately int-encoded PR1 `SafetyTier`/`ReferralCategory` and the existing `CompletionStatus` — PR1's safety enums are left untouched.

## Verification

- Backend: full suite **1395 passed / 0 failed**. New coverage — pure-function `Apply(PlanAdaptedFromLog)` unit tests (restructure swap + meso target; no future-micro synthesis; nudge; absent-micro-week skip; null-After removal skip; empty diff; 1-based-week guard) + `ConversationTurnView` factory tests; integration tests appending restructure/nudge/safety-only to a real stream (projection mutates + turns emitted/not, diff retained, no future-micro synthesis, live-aggregation replay idempotency); a targeted **registration guard** (direct `RegisteredEventTypes` membership + `_v1` suffix); and `ConversationController` integration (anonymous 401, empty cases, adaptation+safety newest-first, cross-runner isolation).
- Frontend: `tsc` + `vite` build clean, **458 Vitest passed**, `npm run lint` 0 errors, OpenAPI **drift gate green** (`codegen:check` exit 0).
- Self-reviewed via an adversarial multi-lens pass (correctness / spec-conventions / cross-file / test-quality) with blind verification; confirmed findings addressed in the third commit.

## Follow-ups found

- **Nullability-aware OpenAPI schema filter (cross-cutting).** `RequireNonNullablePropertiesSchemaFilter` marks every `$ref`-typed property `required` regardless of C# nullability — a **pre-existing, repo-wide** limitation affecting 40+ properties, including the already-shipping nullable `MesoDaySlotOutput.workoutType` (null on Rest days). It is type-accuracy only, not a runtime bug: the generated zod is not applied to responses, and consuming modules hand-write accurate discriminated domain models (`plan/models/plan.model.ts`). PR3 follows the convention and flags the caveat for PR7 in the generated barrel; the structural fix (make the filter consult `NullabilityInfoContext` + regenerate codegen repo-wide) is a separate, reviewed change, not PR3 scope.

## Summary by CodeRabbit

* **New Features**
  * Explain-the-change conversation panel now exposes turn-based messages for plan adaptations and safety alerts.
  * New GET /api/v1/conversation/turns endpoint returns newest-first turns including structured before/after workout and weekly-target diffs and human rationale.

* **Bug Fixes / Validation**
  * Improved validation: malformed adaptation diffs are rejected to prevent incorrect plan mutations.
  * Deterministic ordering for simultaneous turns ensures consistent newest-first presentation.

* **Tests**
  * Added extensive integration and unit tests covering ordering, idempotency, validation, and DTO mapping.


backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationProjection.cs
```diff
@@ -0,0 +1,98 @@
...
+        var turns = view.Turns.ToList();
+        var existing = turns.FindIndex(t => t.TriggeringPlanEventId == turn.TriggeringPlanEventId);
+        if (existing >= 0)
+        {
+            turns[existing] = turn;
+        }
+        else
+        {
+            turns.Add(turn);
+        }
```
leehopper, line 94: **🟠 [HIGH] 🟠 [high] Upsert find-or-replace dedup branch (the actual idempotency guard) has no test coverage**

ConversationProjection.Upsert claims idempotency-under-replay via find-or-replace by TriggeringPlanEventId: if a turn for the event id already exists it overwrites in place (turns[existing] = turn) rather than appending. But every test applies each event exactly once: Marten inline projection delivers each event once, and IdempotentReplay does a single AggregateStreamAsync pass over distinct event ids. So the FindIndex-hit branch (existing >= 0) never executes. The dedup logic that the class XML doc and the spec call the idempotency mechanism is entirely untested; the replay test actually only proves one-turn-per-distinct-event, not that re-applying the same event is non-duplicating.

**Suggested fix:** Add a test that re-applies an event with an already-seen `TriggeringPlanEventId` (e.g. drive a daemon re-projection, or call the projection twice over a stream that already produced the turn) and assert the turn count stays 1 — this is the only path that exercises the `existing >= 0` replace branch the XML doc calls the idempotency guard.

---
*Confidence: 75% · type: test coverage · deep-review*

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationController.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            return Ok(ConversationTurnsResponseDto.Empty);
+        }
+
+        // Conjoined tenancy: the DI session has no tenant, so open a per-request
+        // tenanted session — same pattern as PlanRenderingController.GetCurrent.
+        await using var session = store.LightweightSession(userId.ToString());
+        var log = await session
+            .LoadAsync<ConversationLogView>(currentPlanId.Value, ct)
+            .ConfigureAwait(false);
```
leehopper, line 66: **🟠 [HIGH] 🟠 [high] Tenancy-isolation test never exercises a cross-tenant Marten read; passes trivially because the other runner has no plan**

The endpoints only real tenancy guarantee is that the per-request session tenanted to caller B cannot load a ConversationLogView that lives under tenant A, even if B somehow resolves As planId. The test GetTurns_OtherRunnersPlan_Returns200_WithEmptyTurns does NOT verify this: runner B is seeded with currentPlanId null, so GetTurns returns ConversationTurnsResponseDto.Empty at the EF currentPlanId-is-null guard before any Marten LoadAsync runs. The conjoined-tenant load path, the actual isolation boundary, is never executed in a cross-tenant scenario.

**Suggested fix:** Seed runner B with a **non-null** `CurrentPlanId` pointing at runner A's plan stream, then assert the response is empty — so the tenanted `LightweightSession(B)` / `LoadAsync` actually runs cross-tenant and proves the conjoined-tenant filter (not just the null-plan short-circuit, which `GetTurns_NoActivePlan` already covers).

---
*Confidence: 70% · type: test coverage · deep-review*

backend/src/RunCoach.Api/Infrastructure/MartenConfiguration.cs
```diff
@@ -135,6 +138,16 @@ public static void Apply(StoreOptions opts, string? connectionString = null)
...
 
+        // Conversation read-model (Slice 3 Unit 2, DEC-079) — a second inline
+        // single-stream projection over the same per-user Plan stream. Keyed by
+        // PlanId so the read endpoint resolves it via the runner's active plan id.
+        // Both inline projections receive each appended event in the SAME Marten
+        // transaction as the append: if either Apply throws (e.g. PlanProjection's
+        // 1-based-week guard), the whole append rolls back and neither read model
+        // mutates — the explanation stays atomic with the plan change (DEC-060).
+        opts.Schema.For<ConversationLogView>().Identity(x => x.PlanId);
+        opts.Projections.Add(new ConversationProjection(), ProjectionLifecycle.Inline);
```
leehopper, line 149: **🟠 [HIGH] 🟠 [high] Cross-projection atomic rollback (DEC-060) is asserted in a comment but never tested end-to-end**

MartenConfiguration documents the load-bearing DEC-060 guarantee: both inline projections receive each appended event in the SAME Marten transaction, so if PlanProjection.Apply throws on a malformed (non-positive-week) diff, the whole append rolls back and neither the plan read model nor the ConversationLogView mutates. This is verified only as a pure unit assertion (PlanProjectionTests throws by calling PlanProjection.Apply directly, off Marten). No integration test appends an invalid PlanAdaptedFromLog through a real session and asserts that (a) SaveChangesAsync throws, (b) the plan projection is unchanged, and (c) the conversation log gained no turn. The atomic, two-projection rollback contract is unproven.

**Suggested fix:** Add an integration test that appends an invalid `PlanAdaptedFromLog` (e.g. a non-positive week in the diff) through a real `IDocumentSession` and asserts (a) `SaveChangesAsync` throws, (b) the `PlanProjectionDto` is unchanged, and (c) the `ConversationLogView` gained no turn — proving the two-projection rollback is atomic, not just that `PlanProjection.Apply` throws in isolation.

---
*Confidence: 60% · type: test coverage · deep-review*

backend/src/RunCoach.Api/Modules/Training/Plan/PlanProjection.cs
```diff
@@ -146,4 +147,123 @@ public static void Apply(FirstMicroCycleCreated @event, PlanProjectionDto dto)
...
+        // read model — same fail-loud contract as Apply(MesoCycleCreated, ...).
+        foreach (var change in @event.Diff.WeeklyTargetChanges)
+        {
+            ThrowIfNonPositiveWeek(change.WeekNumber);
+        }
+
+        foreach (var change in @event.Diff.WorkoutChanges)
+        {
+            ThrowIfNonPositiveWeek(change.WeekNumber);
+        }
```
leehopper, line 180: **🟡 [MEDIUM] 🟡 [medium] Non-positive week validation is only tested for WeeklyTargetChanges, not for WorkoutChanges**

Apply(PlanAdaptedFromLog) validates 1-based week numbers in two separate loops — one over WeeklyTargetChanges and one over WorkoutChanges — each calling ThrowIfNonPositiveWeek so a malformed diff fails the whole transaction. The only throwing test (Apply_PlanAdaptedFromLog_NonPositiveWeekNumber_Throws) supplies the invalid week via a WeeklyTargetChange with empty WorkoutChanges, so the WorkoutChanges validation loop is never exercised. A WorkoutChange with WeekNumber 0 or -1 is not tested.

**Suggested fix:** Add a `[Theory]`/`[Fact]` case supplying the invalid week via a `WorkoutChange` (with empty `WeeklyTargetChanges`) so the `WorkoutChanges` validation loop is exercised. Today a regression that deleted that loop would pass the whole suite.

---
*Confidence: 75% · type: test coverage · deep-review*

backend/src/RunCoach.Api/Modules/Training/Plan/PlanProjection.cs
```diff
@@ -146,4 +147,123 @@ public static void Apply(FirstMicroCycleCreated @event, PlanProjectionDto dto)
...
+            var index = weeks.FindIndex(w => w.WeekNumber == change.WeekNumber);
+            if (index >= 0)
+            {
+                weeks[index] = weeks[index] with { WeeklyTargetKm = change.AfterWeeklyTargetKm };
+            }
+
+            // A target change for a week not present in MesoWeeks is skipped — the
+            // meso tier is the authoritative week list; the projection never
+            // synthesizes weeks the plan-generation flow did not emit.
+        }
```
leehopper, line 207: **🟡 [MEDIUM] 🟡 [medium] ApplyWeeklyTargetChanges skip branch (target change for a week absent from MesoWeeks) is untested**

In ApplyWeeklyTargetChanges, when FindIndex returns -1 (a WeeklyTargetChange whose WeekNumber is not present in dto.MesoWeeks) the change is silently skipped — the projection never synthesizes a meso week. Every test feeds week numbers that already exist (2 and 3), so the index < 0 skip branch never executes. The documented non-synthesize contract for weekly targets is unverified.

**Suggested fix:** Add a test feeding a `WeeklyTargetChange` for a positive week **absent** from `MesoWeeks` (e.g. week 5 against the 4-week canonical DTO) and assert `MesoWeeks` is unchanged — locking in the documented non-synthesize contract.

---
*Confidence: 70% · type: test coverage · deep-review*

backend/src/RunCoach.Api/Modules/Training/Plan/PlanProjection.cs
```diff
@@ -146,4 +147,123 @@ public static void Apply(FirstMicroCycleCreated @event, PlanProjectionDto dto)
...
+
+                var workoutIndex = workouts.FindIndex(w => w.DayOfWeek == change.DayOfWeek);
+                if (workoutIndex >= 0)
+                {
+                    workouts[workoutIndex] = change.After;
+                }
+                else
+                {
+                    workouts.Add(change.After);
+                }
```
leehopper, line 250: **🟡 [MEDIUM] 🟡 [medium] ApplyWorkoutChanges add-new-workout branch (DayOfWeek not present in a populated micro week) is untested**

In ApplyWorkoutChanges, when a WorkoutChange targets a populated micro week but its DayOfWeek matches no existing workout (workoutIndex < 0), the code adds a new workout (workouts.Add(change.After)) — the add-a-workout-on-a-rest-day path that WorkoutChange documents as a null-Before addition. Every workout-change test uses DayOfWeek 0, which already exists in week 1, so only the workoutIndex >= 0 replace branch is exercised. The append/add branch is uncovered.

**Suggested fix:** Add a test with a `WorkoutChange` whose `DayOfWeek` is not present in the populated micro week (e.g. a rest day) and assert the workout is appended — the `workouts.Add(change.After)` branch is currently never hit (all tests use `DayOfWeek 0`).

---
*Confidence: 70% · type: test coverage · deep-review*

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationController.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// <returns>200 with the turns (possibly empty); 401 when the user claim is missing or malformed.</returns>
+    [HttpGet("turns")]
+    [ProducesResponseType(typeof(ConversationTurnsResponseDto), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    public async Task<IActionResult> GetTurns(CancellationToken ct)
+    {
+        if (!TryGetUserId(out var userId))
+        {
+            return MissingUserClaim();
+        }
```
leehopper, line 42: **🟡 [MEDIUM] 🟡 [medium] TryGetUserId-false / MissingUserClaim 401 branch is never exercised by a test**

The controller has an in-action authorization branch: when an authenticated principal lacks a parseable NameIdentifier/sub claim, TryGetUserId returns false and the action returns a 401 ProblemDetails via MissingUserClaim(). The only 401 test (GetTurns_Anonymous_Returns401) sends no token at all, which the CookieOrBearer policy rejects at middleware BEFORE the action runs, so it never reaches TryGetUserId. The malformed-subject branch and the ProblemDetails shape it returns (type, title, traceId) are untested.

**Suggested fix:** Mint a validly-signed bearer token with a missing/non-Guid `sub` claim and assert 401 + the `MissingUserClaim` ProblemDetails `type`. The sibling `ClientErrorsController` test (`Report_BearerToken_With_Non_Guid_Sub_Returns_401_...`) is a ready-made template; the only current 401 test is anonymous and is rejected at middleware before the action runs.

---
*Confidence: 65% · type: test coverage · deep-review*

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationController.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        if (log is null)
+        {
+            return Ok(ConversationTurnsResponseDto.Empty);
+        }
+
+        var turns = log.Turns
+            .OrderByDescending(t => t.CreatedAt)
+            .Select(MapTurn)
+            .ToArray();
```
leehopper, line 81: **🟡 [MEDIUM] 🟡 [medium] Newest-first ordering is untested for turns that share an identical CreatedAt (same-transaction appends)**

The read endpoint orders turns purely by OrderByDescending(t => t.CreatedAt) with no secondary tie-break. Events appended in a single Marten transaction share one server timestamp (EventAppendMode.Rich uses transaction_timestamp()), so two turns from one SaveChangesAsync carry equal CreatedAt and their relative order under OrderByDescending is non-deterministic. The ordering test only covers two events appended in SEPARATE transactions (distinct timestamps); the idempotency test appends two events in ONE transaction but asserts only count and id-uniqueness, never order. No test pins a stable newest-first order when timestamps tie.

**Suggested fix:** Append two events in **one** transaction (shared `CreatedAt`) and assert a stable newest-first order. Related: the controller orders only by `CreatedAt`; `OrderByDescending` is a *stable* sort so same-transaction ties silently keep oldest-first (newest-first is violated). Add a deterministic secondary key (e.g. the Marten event `Version`) so the contract is well-defined before PR5 wires co-transaction appends.

---
*Confidence: 55% · type: test coverage · deep-review*

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationTurnView.cs
```diff
@@ -0,0 +1,106 @@
...
+    public required SafetyTier SafetyTier { get; init; }
+
+    /// <summary>Gets the referral category for a safety turn; <see cref="Safety.ReferralCategory.None"/> otherwise.</summary>
+    public required ReferralCategory ReferralCategory { get; init; }
+
+    /// <summary>Gets the adaptation kind for an adaptation turn; null for a safety turn.</summary>
+    public AdaptationKind? AdaptationKind { get; init; }
+
+    /// <summary>Gets the structured before/after diff for an adaptation turn; null for a safety turn.</summary>
+    public PlanAdaptationDiff? Diff { get; init; }
```
leehopper, line 39: **🟡 [MEDIUM] 🟡 [medium] ConversationTurnView encodes role-discriminated data as independent nullable fields, allowing inconsistent turns**

Encapsulation: 5/10, Expression: 4/10, Usefulness: 7/10, Enforcement: 5/10. The record models two distinct turn shapes (AssistantAdaptation vs SystemSafety) as one flat record where EscalationLevel/AdaptationKind/Diff are nullable and SafetyTier/ReferralCategory are always present, with the Role/payload correlation enforced only by the two static factories FromAdaptation/FromSafety. Because every property is a public init setter, any caller can construct an illegal combination (e.g. Role=SystemSafety with a non-null Diff, or Role=AssistantAdaptation with a null Diff) that no factory would ever produce; the type itself does not exclude it. The read-model and DTO both inherit this, so the frontend switch on Role cannot rely on the type system to guarantee the matching fields are populated. A Marten-persisted JSONB document plus OpenAPI codegen makes a full discriminated-union refactor genuinely constrained, so this is a design weakness rather than a clear bug. Recommendation: make the construction-only invariant harder to bypass — e.g. a private parameterless ctor / non-public init surface so the factories are the sole entry point, or document the invariant and add a guard in the factories, so an adaptation turn cannot be built without its Diff and a safety turn cannot carry one.

**Suggested fix:** Make the two static factories the sole construction path — e.g. a non-public `init`/parameterless ctor — and/or split into `AdaptationTurnView`/`SafetyTurnView`, so an adaptation turn cannot exist without its `Diff` and a safety turn cannot carry one. (Marten JSONB + OpenAPI codegen constrain a full discriminated-union refactor, so a guarded factory may be the pragmatic floor.)

---
*Confidence: 25% · type: type design · deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/WorkoutChange.cs
```diff
@@ -0,0 +1,23 @@
...
+/// </summary>
+/// <param name="WeekNumber">The 1-based plan week the workout belongs to.</param>
+/// <param name="DayOfWeek">The day of week, 0=Sunday..6=Saturday.</param>
+/// <param name="Before">The workout before the adaptation, or null when added.</param>
+/// <param name="After">The workout after the adaptation, or null when removed.</param>
+public sealed record WorkoutChange(
+    int WeekNumber,
+    int DayOfWeek,
+    WorkoutOutput? Before,
+    WorkoutOutput? After);
```
leehopper, line 23: **💡 [LOW] 💡 [low] WorkoutChange permits (Before=null, After=null), a meaningless no-op change the type does not exclude**

Encapsulation: 6/10, Expression: 5/10, Usefulness: 7/10, Enforcement: 4/10. WorkoutChange documents that a null After means a removed workout and a null Before means an added one, but both Before and After are independently nullable, so the (null, null) combination — a change that adds, removes, and edits nothing — is representable. The XML doc encodes the add/remove/edit semantics in prose only; nothing constructs the invalid pair, but the type permits it and the projection silently no-ops on it (ApplyWorkoutChanges skips when After is null), so a malformed diff would be absorbed without signal rather than rejected. Impact is low (this slice only produces edit changes with both populated), but since the diff is the structured contract the panel renders, an unconstrained no-op entry is a latent expressiveness gap. Recommendation: if removal is not modeled this slice (as the projection comment states), require After non-null; if both directions are needed later, document the (null,null) case as invalid and guard it at construction.

**Suggested fix:** If removal isn't modeled this slice (as the projection comment states), make `After` non-null; if both add and remove are needed later, document `(null, null)` as invalid and guard it at construction so a malformed diff is rejected rather than silently absorbed.

---
*Confidence: 20% · type: type design · deep-review*

leehopper (commented):
### Deep Review — PR #163

**10 improvement suggestions** (8 test coverage, 1 type design medium, 1 type design low) — **no blocking issues, no confirmed runtime bugs.** Security, cross-file impact, conventions/intent, and simplification all came back clean; the production code matches established repo patterns and all spec BDD criteria.

The theme is a gap between what the code guarantees and what the tests exercise: several load-bearing branches (the projection idempotency dedup, the cross-tenant read, the DEC-060 atomic rollback, the skip/add branches) pass trivially or are never hit. All 10 findings survived an independent blind-challenge round (Frontier/all-Opus, scores 85–98).

One related latent issue was filtered out: `ConversationController` orders turns only by `CreatedAt`, so same-transaction appends (which PR5 will produce) tie and lose newest-first ordering — see **test-3** for the fix.

Each finding is posted as an inline comment below.

---
Generated by deep-review | Reviewed up to: c1b6427fe269f2e2736cc0dae87bf80ff758f7b6

backend/tests/RunCoach.Api.Tests/Modules/Training/Plan/PlanProjectionTests.cs
```diff
@@ -293,6 +295,193 @@ public void FullSequence_PlanGeneratedFourMesoFirstMicro_LandsCanonicalShape()
...
+            SafetyTier.Green,
+            "n/a",
+            diff);
+
+        // Act
+        var act = () => PlanProjection.Apply(adaptation, actualDto);
+
+        // Assert
+        act.Should().Throw<ArgumentOutOfRangeException>();
+    }
```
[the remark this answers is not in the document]
leehopper: Addressed in e129c87, with one refinement over the suggested diff. Combining a `WorkoutChange` **and** a `WeeklyTargetChange` into the existing theory means the (earlier) `WeeklyTargetChanges` loop throws first — so a regression that removed *only* the `WorkoutChanges` loop would still be masked. Instead added a dedicated `Apply_PlanAdaptedFromLog_NonPositiveWeekNumber_InWorkoutChange_Throws` `[Theory]` that carries the invalid week on a `WorkoutChange` with **empty** `WeeklyTargetChanges`, so the `WorkoutChanges` validation loop is exercised in isolation.

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationProjection.cs
leehopper: Fixed in e129c87. Added a `ConversationProjectionTests` unit class that applies the same `IEvent<PlanAdaptedFromLog>` id twice against one view and asserts the single turn is **replaced in place** (count stays 1, content is the second payload). A single-pass aggregation / inline delivery only ever sees each id once, so this is the only path that reaches the `existing >= 0` find-or-replace branch.

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationController.cs
leehopper: Fixed in e129c87. `GetTurns_OtherRunnersPlan` now seeds runner B with **runner A's** plan id, so the request passes the EF `currentPlanId is null` guard and actually runs the tenanted Marten `LoadAsync` against A's stream id — conjoined tenancy (the real isolation boundary) returns null → empty. The null-plan short-circuit stays covered separately by `GetTurns_NoActivePlan`.

backend/src/RunCoach.Api/Infrastructure/MartenConfiguration.cs
leehopper: Fixed in e129c87. Added `InvalidPlanAdaptedFromLog_RollsBackBothInlineProjections_Atomically`: it appends a non-positive-week diff through a real `IDocumentSession`, asserts `SaveChangesAsync` throws, and then asserts that **both** the `PlanProjectionDto` and the `ConversationLogView` are unchanged (log still has zero turns). That proves the two-projection atomic rollback end-to-end, not just that `PlanProjection.Apply` throws off-Marten in isolation.

backend/src/RunCoach.Api/Modules/Training/Plan/PlanProjection.cs
leehopper: Fixed in e129c87 (same fix as the CodeRabbit thread on `PlanProjectionTests`). Added `Apply_PlanAdaptedFromLog_NonPositiveWeekNumber_InWorkoutChange_Throws` — the invalid week rides a `WorkoutChange` with empty `WeeklyTargetChanges`, so the `WorkoutChanges` validation loop is covered independently and a regression deleting just that loop now fails.

backend/src/RunCoach.Api/Modules/Training/Plan/PlanProjection.cs
leehopper: Fixed in e129c87. Added `Apply_PlanAdaptedFromLog_WeeklyTargetChangeForAbsentWeek_LeavesMesoWeeksUnchanged` — a target change for week 5 (absent from the 4-week canonical DTO) leaves `MesoWeeks` byte-for-byte unchanged, locking in the documented non-synthesize contract (the `FindIndex < 0` skip branch).

backend/src/RunCoach.Api/Modules/Training/Plan/PlanProjection.cs
leehopper: Fixed in e129c87. Added `Apply_PlanAdaptedFromLog_WorkoutChangeForAbsentDay_AddsWorkoutToMicroWeek` — a `WorkoutChange` on day 3 (absent from the populated week-1 micro) is appended rather than replacing, exercising the `workouts.Add(change.After)` branch (every prior workout-change test used day 0, which already existed).

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationController.cs
leehopper: Fixed in e129c87. Added `GetTurns_BearerToken_With_Non_Guid_Sub_Returns_401_Problem` plus a `MintBearerTokenWithRawSub` helper (mirrors the `PlanRenderingController` template): a validly-signed token with a non-Guid `sub` passes the CookieOrBearer middleware, fails the in-action `TryGetUserId`, and returns the `MissingUserClaim` 401 with `application/problem+json`.

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationController.cs
leehopper: Fixed in e129c87 — including the code change you flagged. `ConversationTurnView` now carries the source event's per-stream Marten `EventVersion`, and the endpoint orders newest-first by `CreatedAt` **then** `EventVersion` (`ThenByDescending`). Added `GetTurns_TurnsSharingCreatedAt_OrderedByEventVersionDescending`: two events appended in one transaction share a `transaction_timestamp()` `CreatedAt`, and the later-appended (higher-version) turn deterministically sorts first — a `CreatedAt`-only stable sort would otherwise keep oldest-first and violate the contract. Doing it now, before PR5 wires co-transaction appends.

backend/src/RunCoach.Api/Modules/Coaching/Conversation/ConversationTurnView.cs
leehopper: Verified and deliberately **not changed** in PR3 — the suggested hardening is constrained on this type:

- **Non-public `init` / parameterless ctor would break Marten rehydration.** The persisted `ConversationLogView.Turns` is deserialized by Marten's serializer, which sets the `init` accessors; making them non-public stops the document round-tripping.
- **A full `AdaptationTurnView`/`SafetyTurnView` split (discriminated union) is constrained** by the Marten JSONB document identity and the OpenAPI codegen the sibling DTO feeds — as the finding itself notes.
- **The two static factories are the only producers**, and both emit valid Role/payload combos; no call site constructs an illegal turn, so a factory guard would be unreachable dead code.

The Role↔payload invariant stays documented on the type. Resolving as a deliberate no-change.

backend/src/RunCoach.Api/Modules/Training/Adaptation/WorkoutChange.cs
leehopper: Verified and deliberately **not changed** in PR3 — the dual-nullability is intentional forward design, not an oversight:

- A null `After` (removal) and null `Before` (addition) are the documented add/remove/edit directions. Making `After` non-null would contradict the removal direction the type is explicitly shaped for in a later slice.
- `WorkoutChange` is persisted inside `PlanAdaptedFromLog.Diff` in the event store and is produced by LLM structured output. A throwing constructor guard would turn today's benign projection no-op (`ApplyWorkoutChanges` already skips a null `After`) into a **hard rehydration failure of a historical event** — a strictly worse failure mode for an event-sourced record.

Impact is low (this slice only emits edits with both populated). Resolving as a deliberate no-change; when removal lands, the `(null, null)` case gets documented-invalid and guarded at the producer rather than on the persisted record.

PR #165, opened 2026-06-08
feat(slice-3): deterministic deviation engine + escalation classifier (PR2)

leehopper:
## Slice 3 — PR2: deterministic deviation engine + escalation classifier (Unit 1)

The pure-C# **decision layer** of the adaptation loop. Given a logged workout and its
frozen prescription snapshot, it decides *how big a deal* a deviation is — without any
LLM, Marten append, or DB access. PR5 (orchestration) wires this into the post-log-create
flow and owns the event append.

Spec: `docs/specs/17-spec-slice-3-adaptation/` Unit 1 + `deterministic-deviation-and-escalation.feature`.
Merge order: PR1 ✅ (#161) → PR3 ✅ (#163) → **PR2 (this)** → PR4 → PR5 → PR6 → PR7.

### What's in it

- **`PaceDerivation`** — shared divide-by-zero-guarded `Distance`/`Duration` → sec/km util,
  lifted out of the presentation-layer `RecentLogFormatter.FormatPace`, which is refactored
  to consume it with **byte-identical output** (guarded by the existing `RecentLogFormatterTests`
  + `ContextAssemblerTests`).
- **`DeviationEngine` + `DeviationResult` + `PaceBandMembership`** — compares actuals vs the
  frozen `WorkoutPrescriptionSnapshot`. Pace deviation is reported as **band membership**
  (inside / faster-than-fast / slower-than-slow + signed sec/km magnitude), distance/duration
  as signed percentages. A `null` snapshot (off-plan) ⇒ `null` (no-op). Skipped or
  zero/near-zero logs ⇒ `Unknown` pace, never a spurious deviation, never a throw.
- **`EscalationClassifier` + `AdaptationSignalState` + `PlanState` + `EscalationDecision`** —
  resolves the DEC-012 level (L0 absorb / L1 micro-adjust / L2 restructure) from the deviation,
  the `SafetyTier`, and the prior signal state, with **asymmetric enter/exit hysteresis** so a
  restructure can't flip-flop (a dead-zone holds until both the cooldown elapses and the rolling
  signal clears the lower exit threshold). A pure transition function.
- **`MicroAdjustPlanner`** — the deterministic L1 swap: moves a missed key workout **forward**
  (DEC-012 — never backward) to the nearest later easy day such that **no two key workouts land
  on consecutive days**; returns `false` (PR5 escalates L1→L2) when no valid swap exists.
- **`AdaptationThresholds`** — named first-pass constants with a `PolicyVersion`, documented as
  pending Unit 6 eval calibration (mirrors the `SafetyKeywordCatalog` resource convention).
- DI: `IDeviationEngine` / `IEscalationClassifier` registered as singletons (mirroring `ISafetyGate`).

Reuses the `EscalationLevel` / `AdaptationKind` / `PlanAdaptationDiff` / `WorkoutChange` types
already shipped by PR3 (#163) — this PR only consumes/produces them.

### Deliberate scope boundary (PR2 vs PR5)

- **State persistence is deferred to PR5.** `AdaptationSignalState` is a pure value threaded in
  and out of the classifier; whether it's persisted as a Marten read-model or recomputed from the
  recent-log window is the spec's explicit open question (DEC-078), resolved when PR5 wires the loop.
- **Red/Amber handling is deferred.** Under `Red` the classifier defensively returns Absorb (the
  real crisis short-circuit is PR5); `Amber` resolves the level like `Green` (the non-increase clamp
  is PR4's post-LLM validator).
- **The swap is computed, not applied.** `MicroAdjustPlanner` produces the `PlanAdaptationDiff`;
  PR5 reads the live micro week and appends the `PlanAdaptedFromLog`.

### Testing

39 new unit tests (`PaceDerivation`, `DeviationEngine`, `EscalationClassifier`, `MicroAdjustPlanner`)
covering every `deterministic-deviation-and-escalation.feature` scenario: within-band absorb,
reschedulable miss ⇒ L1, 2–3 minor ⇒ L1, sustained / 3+ consecutive missed ⇒ L2, off-plan no-op,
over-performance no-upgrade, the hysteresis dead-zone, divide-by-zero edges, band-membership
reporting, and the no-consecutive-key swap invariant. Built TDD (red → green). **Full suite
1443/1443 green.** Adversarially reviewed pre-commit (bug / conventions / cross-file lenses + blind
verification); the two confirmed findings (both doc/naming accuracy) are folded in.

### Follow-ups found

- _None blocking._ PR5 consumes the planner + classifier and decides the signal-state persistence
  mechanism (the deferred open question above).

## Summary by CodeRabbit

* **New Features**
  * Intelligent workout deviation detection and escalation (micro-adjusts or restructures plans).
  * Smart forward-rescheduling for missed key workouts that prevents consecutive key sessions.
  * Improved pace derivation used when rendering recent workout pace (returns no pace when it can’t be derived).

* **Tests**
  * Expanded unit test coverage for deviation detection, escalation logic, micro-adjust planning, and pace derivation.

### DIFF
--- a/backend/src/RunCoach.Api/Modules/Training/Adaptation/AdaptationSignalState.cs
+++ b/backend/src/RunCoach.Api/Modules/Training/Adaptation/AdaptationSignalState.cs
@@ -1,0 +1,31 @@
+namespace RunCoach.Api.Modules.Training.Adaptation;
+
+/// <summary>
+/// The minimal per-user signal + plan-state the deterministic escalation classifier
+/// threads between evaluations (Slice 3 PR2 / Unit 1, DEC-078). Pure data: the
+/// classifier consumes a prior state and returns the next one. <strong>How this state
+/// is persisted (a small Marten read-model vs. recomputed per evaluation from the
+/// recent-log window) is deliberately left to the PR5 orchestration</strong> — the
+/// spec's open question — so the deterministic layer stays a pure, total transition.
+/// No EWMA/ACWR/CTL/ATL/TSB load model is computed at MVP-0.
+/// </summary>
+/// <param name="PlanState">The current plan-state in the hysteresis machine.</param>
+/// <param name="RollingDeviationScore">
+/// A simple accumulator of recent under-performance: it steps up on an under-performing
+/// log and decays toward zero on an on-target log. Not a calibrated load metric.
+/// </param>
+/// <param name="ConsecutiveMissedDays">Run of consecutive skipped days; resets on any non-skipped log.</param>
+/// <param name="LastAdaptationOn">
+/// The day the most recent restructure fired, used for the cooldown half of the
+/// hysteresis; <c>null</c> until the first restructure.
+/// </param>
+public sealed record AdaptationSignalState(
+    PlanState PlanState,
+    double RollingDeviationScore,
+    int ConsecutiveMissedDays,
+    DateOnly? LastAdaptationOn)
+{
+    /// <summary>Gets the starting state for a user with no adaptation history.</summary>
+    public static AdaptationSignalState Initial { get; } =
+        new(PlanState.OnTrack, 0.0, 0, null);
+}
--- a/backend/src/RunCoach.Api/Modules/Training/Adaptation/EscalationClassifier.cs
+++ b/backend/src/RunCoach.Api/Modules/Training/Adaptation/EscalationClassifier.cs
@@ -1,0 +1,103 @@
+using RunCoach.Api.Modules.Training.Safety;
+using RunCoach.Api.Modules.Training.Workouts;
+
+namespace RunCoach.Api.Modules.Training.Adaptation;
+
+/// <summary>
+/// Deterministic escalation classifier (Slice 3 PR2 / Unit 1, DEC-012/DEC-078):
+/// resolves L0 absorb / L1 micro-adjust / L2 restructure from a
+/// <see cref="DeviationResult"/>, the safety tier, and a prior signal state, applying
+/// asymmetric enter/exit hysteresis so a restructure cannot flip-flop. Pure and
+/// stateless — no LLM, no I/O; the L2 restructure is where PR5 hands off to the LLM.
+/// </summary>
+public sealed class EscalationClassifier : IEscalationClassifier
+{
+    /// <inheritdoc />
+    public EscalationDecision Classify(
+        DeviationResult deviation,
+        SafetyTier safetyTier,
+        AdaptationSignalState priorState)
+    {
+        ArgumentNullException.ThrowIfNull(deviation);
+        ArgumentNullException.ThrowIfNull(priorState);
+
+        // Red short-circuits to safety-only in the PR5 orchestration; defensively the
+        // classifier yields no plan adaptation and leaves the signal state untouched.
+        if (safetyTier == SafetyTier.Red)
+        {
+            return Absorb(priorState);
+        }
+
+        var missed = deviation.CompletionStatus == CompletionStatus.Skipped;
+        var newStreak = missed ? priorState.ConsecutiveMissedDays + 1 : 0;
+        var newScore = IsUnderPerformance(deviation)
+            ? Math.Min(
+                priorState.RollingDeviationScore + AdaptationThresholds.MinorDeviationStep,
+                AdaptationThresholds.MaxRollingDeviationScore)
+            : priorState.RollingDeviationScore * AdaptationThresholds.OnTargetDecayFactor;
+
+        // Asymmetric hysteresis: once a restructure has fired, suppress re-firing until
+        // the cooldown has elapsed AND the signal has cleared the (lower) exit threshold.
+        if (priorState.PlanState == PlanState.NeedsAdjustment)
+        {
+            var cooldownActive = priorState.LastAdaptationOn is { } last
+                && deviation.OccurredOn.DayNumber - last.DayNumber < AdaptationThresholds.RestructureCooldownDays;
+            if (cooldownActive || newScore > AdaptationThresholds.RestructureExitScore)
+            {
+                // Still inside the dead-zone — absorb without re-firing, hold the state.
+                return Absorb(priorState with
+                {
+                    RollingDeviationScore = newScore,
+                    ConsecutiveMissedDays = newStreak,
+                });
+            }
+
+            // Cleared: cooldown elapsed and the signal settled at or below the exit threshold.
+            return Absorb(new AdaptationSignalState(
+                PlanState.OnTrack, newScore, newStreak, priorState.LastAdaptationOn));
+        }
+
+        // L2 restructure: sustained under-performance crossing the enter threshold, or a
+        // run of consecutive missed days. This is where PR5 hands off to the LLM.
+        if (newScore >= AdaptationThresholds.RestructureEnterScore
+            || newStreak >= AdaptationThresholds.ConsecutiveMissedDaysForRestructure)
+        {
+            return new EscalationDecision(
+                EscalationLevel.Restructure,
+                AdaptationKind.Restructure,
+                new AdaptationSignalState(PlanState.NeedsAdjustment, newScore, newStreak, deviation.OccurredOn));
+        }
+
+        // L1 micro-adjust: a single reschedulable missed/moved key workout, or accumulated
+        // minor deviations reaching the micro-adjust threshold.
+        if ((missed && deviation.IsKeyWorkout) || newScore >= AdaptationThresholds.MicroAdjustEnterScore)
+        {
+            return new EscalationDecision(
+                EscalationLevel.MicroAdjust,
+                AdaptationKind.Nudge,
+                new AdaptationSignalState(PlanState.MinorDeviation, newScore, newStreak, priorState.LastAdaptationOn));
+        }
+
+        // L0 absorb.
+        var nextPlanState = newScore >= AdaptationThresholds.MinorDeviationStep
+            ? PlanState.MinorDeviation
+            : PlanState.OnTrack;
+        return new EscalationDecision(
+            EscalationLevel.Absorb,
+            AdaptationKind.Absorb,
+            new AdaptationSignalState(nextPlanState, newScore, newStreak, priorState.LastAdaptationOn));
+    }
+
+    private static EscalationDecision Absorb(AdaptationSignalState state) =>
+        new(EscalationLevel.Absorb, AdaptationKind.Absorb, state);
+
+    /// <summary>
+    /// A log under-performs when it was not completed (skipped or cut short), its pace
+    /// was slower than the prescribed band, or it ran meaningfully short of the
+    /// prescribed distance. Over-performance and on-target both fall through to decay.
+    /// </summary>
+    private static bool IsUnderPerformance(DeviationResult deviation) =>
+        deviation.CompletionStatus != CompletionStatus.Complete
+        || deviation.PaceBand == PaceBandMembership.SlowerThanSlow
+        || deviation.DistanceDeviationPercent < -AdaptationThresholds.DistanceTolerancePercent;
+}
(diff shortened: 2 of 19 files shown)


leehopper (commented):
### Deep Review

Reviewed the deterministic deviation engine + escalation classifier (pure decision layer; no LLM/DB/I/O). **Security, cross-file impact, and code-simplification came back clean**, and conventions/comment-accuracy are sound — the hysteresis math and the no-consecutive-key swap trace correctly against the BDD spec.

Posting **4 findings** (1 low-value `OverPerformanceCapPercent`-unused note skipped as documented-intentional):

- **2 latent design notes for PR5** (logically confirmed by blind challenge; not reachable until PR5 wires orchestration, so flagged now rather than blocking): the `NeedsAdjustment` dead-zone swallowing fresh L1/L2 triggers, and `AdaptationSignalState` not enforcing its own invariants ahead of PR5's rehydration path.
- **2 test-coverage gaps**: `CompletionStatus.Partial` untested end-to-end, and the −5% distance under-performance branch never independently exercised.

Pipeline: 7 Opus discovery agents → 6 raw findings → 5 verified (1 eliminated: bad line ref) → Sonnet validation → 5-finding blind challenge (Opus). No security or injection findings.

Generated by deep-review | Reviewed up to: b1cf125a0d22ee0b96fd5d76882042d232df00a3

backend/src/RunCoach.Api/Modules/Training/Adaptation/EscalationClassifier.cs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    RollingDeviationScore = newScore,
+                    ConsecutiveMissedDays = newStreak,
+                });
+            }
+
+            // Cleared: cooldown elapsed and the signal settled at or below the exit threshold.
+            return Absorb(new AdaptationSignalState(
+                PlanState.OnTrack, newScore, newStreak, priorState.LastAdaptationOn));
+        }
```
leehopper, line 58: **🟠 [HIGH] NeedsAdjustment dead-zone swallows fresh L1/L2 triggers indefinitely**

While `priorState.PlanState == NeedsAdjustment`, **both** arms of this branch `return Absorb(...)` before the L2 restructure check (line 62) and the L1 key-miss micro-adjust check (line 73), so those triggers are structurally unreachable in this state.

Because `newScore` only ever **rises** on under-performance (`Math.Min(... + MinorDeviationStep, Max=3.0)`) and never decays except on on-target logs, the dead-zone's exit condition (`!cooldownActive && newScore <= RestructureExitScore=1.0`) can never be met while the runner keeps under-performing. A runner already in `NeedsAdjustment` who then accrues a **brand-new** 3-consecutive-missed-day streak, or misses a key workout, receives only `Absorb` decisions — **not just for the 7-day cooldown, but indefinitely** — with no signal that an adaptation was suppressed. The only exit is *recovery* (on-target logs decaying the score below 1.0), i.e. precisely the runner who needs no adaptation.

The blind-challenge round scored this low **only on reachability** (`Classify` has no production caller yet — orchestration is deferred to PR5), not on correctness: the validator and challenger both agreed the logic is accurate. Flagging it now because PR5 will make this state reachable.

**Suggested fix:** Decide explicitly whether suppression inside the dead-zone should apply to the low-cost L1 micro-adjust at all (the spec's anti-flip-flop guarantee is about L2 restructure). Consider either (a) still allowing L1 key-miss swaps while in `NeedsAdjustment`, and/or (b) honoring the `ConsecutiveMissedDaysForRestructure` hard trigger as a dead-zone override so a genuinely deteriorating runner still escalates once the cooldown elapses, and/or (c) emitting an observability signal when a trigger is suppressed. Add a test that drives a fresh hard trigger while already in `NeedsAdjustment` post-cooldown to lock in whichever behavior is intended.

backend/src/RunCoach.Api/Modules/Training/Adaptation/AdaptationSignalState.cs
```diff
@@ -0,0 +1,31 @@
...
+/// </param>
+public sealed record AdaptationSignalState(
+    PlanState PlanState,
+    double RollingDeviationScore,
+    int ConsecutiveMissedDays,
+    DateOnly? LastAdaptationOn)
+{
+    /// <summary>Gets the starting state for a user with no adaptation history.</summary>
+    public static AdaptationSignalState Initial { get; } =
+        new(PlanState.OnTrack, 0.0, 0, null);
```
leehopper, line 30: **🟡 [MEDIUM] AdaptationSignalState enforces none of its hysteresis invariants**

This positional record is the threaded state of the asymmetric-hysteresis machine, but enforces none of the cross-field invariants `EscalationClassifier` relies on. The ctor accepts a negative `RollingDeviationScore` or `ConsecutiveMissedDays`, a score above `MaxRollingDeviationScore`, and — most consequentially — `(NeedsAdjustment, …, null LastAdaptationOn)`, which silently disables the cooldown half of the hysteresis (the cooldown branch reads `LastAdaptationOn` only when non-null, so a null there makes `cooldownActive=false`).

Harmless today — the only producer is the classifier, which always emits valid states. But **PR5's rehydrate-from-persistence path (DEC-078) is exactly where an invalid state could enter**, and that path will reconstruct this type directly. The blind challenger confirmed the missing-enforcement observation is factually accurate; it scored low only because no such caller exists yet.

**Suggested fix:** Mirror the same-module `WorkoutPrescriptionSnapshot.Create()` pattern — a validating factory (or guarded ctor) that clamps the score to `[0, MaxRollingDeviationScore]`, rejects a negative streak, and rejects `NeedsAdjustment` paired with a null `LastAdaptationOn` — so PR5's deserialization boundary cannot reintroduce an invalid transition state.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EscalationClassifier.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// <summary>
+    /// A log under-performs when it was not completed (skipped or cut short), its pace
+    /// was slower than the prescribed band, or it ran meaningfully short of the
+    /// prescribed distance. Over-performance and on-target both fall through to decay.
+    /// </summary>
+    private static bool IsUnderPerformance(DeviationResult deviation) =>
+        deviation.CompletionStatus != CompletionStatus.Complete
+        || deviation.PaceBand == PaceBandMembership.SlowerThanSlow
+        || deviation.DistanceDeviationPercent < -AdaptationThresholds.DistanceTolerancePercent;
```
leehopper, line 102: **🟡 [MEDIUM] CompletionStatus.Partial is untested across the whole decision layer**

The decision layer has three `CompletionStatus` values (`Complete`, `Partial`, `Skipped`) but no test in the Adaptation/Computations folders ever exercises `Partial`. Two `Partial`-specific behaviors are therefore unverified:

1. In `IsUnderPerformance` (line 100), `CompletionStatus != Complete` means a `Partial` log accumulates the rolling deviation score **exactly like a Skipped one**.
2. In `DeviationEngine.ClassifyPace`, only `Skipped` short-circuits to `Unknown` pace — so a `Partial` log still **derives a real pace** from its (possibly truncated) distance/duration, which can then independently drive the pace-band and distance triggers.

A regression narrowing `!= Complete` to `== Skipped` would silently stop escalating runners who repeatedly cut key sessions short, and no test would fail.

**Suggested fix:** Add classifier + engine cases that pass `CompletionStatus.Partial` (e.g. a Partial log that derives an in-band pace but a short distance) to pin down both the score-accumulation and pace-derivation behavior.

leehopper, line 102: **💡 [LOW] Distance-shortfall under-performance branch never independently tested**

`IsUnderPerformance` has three independent OR clauses; the third — `DistanceDeviationPercent < -DistanceTolerancePercent` (−5%) — is never the **sole** driver in any test. Every `MinorUnder` helper fires via `SlowerThanSlow` pace (with `distancePct` left at its 0.0 default), and the only large negative distances (−100) appear on `Skipped` logs where the not-complete clause already fires. So a `Complete`, `InsideBand` log that merely ran well short of prescribed distance is never tested, and the −5% tolerance boundary is unverified — a sign flip or tolerance change on this clause would fail no test.

**Suggested fix:** Add a case with a `Complete`, `InsideBand` log whose `DistanceDeviationPercent` sits just beyond −5% (and a paired just-inside case) to exercise the boundary.

leehopper: Fixed in c6edfcf. `CompletionStatus.Partial` is now exercised across both layers:

- `EscalationClassifierTests.Classify_PartialLogInBand_UnderPerformsButDoesNotCountAsMissedDay` — a Partial in-band log under-performs via the not-Complete clause (score steps by `MinorDeviationStep`) yet `ConsecutiveMissedDays` stays 0, since only a Skipped log advances the streak.
- `EscalationClassifierTests.Classify_TwoPartialLogs_AccumulateToMicroAdjust` — two Partial logs accumulate to a micro-adjust, pinning the score-accumulation behavior.
- `DeviationEngineTests.Evaluate_PartialLog_DerivesRealPaceFromTruncatedActualsAndCarriesPartialStatus` — a Partial derives a real `SlowerThanSlow` band from its truncated actuals (only Skipped short-circuits to `Unknown`) and carries `Partial` status.

A regression narrowing `!= Complete` to `== Skipped`, or short-circuiting Partial pace to `Unknown`, now fails a test.

leehopper: Fixed in c6edfcf. `EscalationClassifierTests.Classify_DistanceShortfallAloneAtToleranceBoundary_DrivesUnderPerformance` is a `[Theory]` over a `Complete`, `InsideBand` log whose only deviation is distance, isolating the third OR clause of `IsUnderPerformance` at the boundary:

- `-6%` → under-performs as the sole driver: score steps to 1.0, `PlanState` becomes `MinorDeviation`.
- `-4%` → within the ±5% tolerance: absorbed as on-target, score stays 0.0, `PlanState` `OnTrack`.

A sign flip or tolerance change on the distance clause now fails.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EscalationClassifier.cs
leehopper: Verified — the analysis is accurate: while `priorState.PlanState == NeedsAdjustment`, both arms return `Absorb` before the L2/L1 checks, and the only dead-zone exit is score decay (recovery), so a still-deteriorating runner stays on `Absorb`. Deferring the behavior change to PR5 rather than acting now, for two reasons:

1. **The current code is spec-compliant.** The PR2 spec scenario *"Asymmetric hysteresis blocks re-firing until the signal clears the exit threshold"* makes clearing the exit threshold the deliberate sole exit, and the anti-flip-flop guarantee is explicitly L2-only. The suggested hard-trigger override would contradict the spec as written, so it's a spec change, not a bug fix.
2. **No caller until PR5, thresholds uncalibrated.** `Classify` is unreferenced until PR5 wires orchestration, and the threshold set is first-pass pending Unit 6 eval calibration against the five `TestProfiles`. The three options you list — (a) let L1 key-miss swaps survive the dead-zone, (b) honor the consecutive-missed hard trigger as a post-cooldown override, (c) emit a suppressed-trigger signal — are genuine product decisions best made with eval data, and (c) specifically can't land in PR2 without breaking this class's pure/no-I/O contract.

Tracked under the PR5 section of the slice pr-strategy (carryover item 1), including the test driving a fresh hard trigger while already in `NeedsAdjustment` post-cooldown to lock in whichever behavior is chosen.

backend/src/RunCoach.Api/Modules/Training/Adaptation/AdaptationSignalState.cs
leehopper: Verified — `AdaptationSignalState` enforces none of its cross-field invariants, and `(NeedsAdjustment, …, null LastAdaptationOn)` does silently disable the cooldown half (the cooldown branch reads `LastAdaptationOn` only when non-null). Deferring the validating factory to PR5:

The only producer today is the classifier, which always emits valid states, so a `Create` guard added now would be unused code guarding a boundary that doesn't exist yet (YAGNI). The spec explicitly defers persistence/rehydration (DEC-078) to PR5 — that rehydrate-from-persistence path is the deserialization boundary where an invalid state could actually enter. The `WorkoutPrescriptionSnapshot.Create()`-mirroring guard (clamp score to `[0, MaxRollingDeviationScore]`, reject a negative streak, reject `NeedsAdjustment` paired with a null `LastAdaptationOn`) belongs there so it ships alongside the caller it protects.

Tracked under the PR5 section of the slice pr-strategy (carryover item 2).

leehopper:
@coderabbitai full review

PR #167, opened 2026-06-09
feat(slice-3): adaptation structured output + prompt + DEC-073 failure shapes (PR4)

leehopper:
## Slice 3 PR4 — adaptation structured output + prompt + DEC-073 failure shapes (Unit 4)

The LLM-facing layer of the adaptation loop. Per the [PR strategy](docs/specs/17-spec-slice-3-adaptation/pr-strategy.md), this lands **behind unreferenced contracts** — the orchestration handler that actually drives a restructure call is PR5. **No live LLM call is added**; everything here is deterministic or structured-output shape work, green in CI without a funded key.

Stacks on PR1 (#161 SafetyGate), PR3 (#163 events/projection), PR2 (#165 deviation engine). Reuses their `AdaptationKind` / `SafetyTier` / `ReferralCategory` enums and `PlanAdaptedFromLog` / `PlanAdaptationDiff` shapes rather than re-defining them.

### What's in scope

- **`PlanAdaptationOutput`** — the net-new Pattern-B (DEC-058) LLM output, frozen at startup via `AdaptationSchema` (mirrors `OnboardingSchema`: `JsonSchemaHelper.GenerateSchema<T>` → `AnthropicSchemaSanitizer`, so no forbidden keyword can ship). `AdaptationKind` + `SafetyTier` discriminators, nullable `NudgePatch` / `RestructurePlan` slots reusing the int-unit `WorkoutOutput`, `NetLoadDelta`, `Rationale`, nullable `ReferralCategory`.
- **`PlanAdaptationOutputValidator`** — enforces what constrained decoding cannot: exactly-one-slot-matches-discriminator (absorb fills neither), **GATE-BEFORE-INCREASE** (`tier != Green ⇒ NetLoadDelta ≤ 0`), and a forward path for any load-reducing restructure.
- **`adaptation.v1.yaml`** — versioned prompt registered as the `adaptation` active version, consuming the assembler context + `DeviationResult` + `EscalationLevel`, Daniels-Gilbert / pace-zone terminology only, recent-log notes framed in `SECTION_NAME` spotlighting delimiters (R-068/DEC-059). Also **neutralized the dead `ACWR` placeholder** in `context-injection-v1.yaml` (no fake load math at MVP-0, DEC-078) — confirmed unread by any code, so zero eval drift.
- **DEC-073 (first live)** — `Transient`/`PermanentCoachingLlmException` translated inside `ClaudeCoachingLlm`; flat `Kind=Error` envelope (`AdaptationResponseDto`) for terminal failures over HTTP-200.

### DEC-073 — SDK reality vs the research sketch

The brainstorm flagged the SDK 12.24.1 failure surface for impl-time verification. Decompiling the installed assembly corrected three load-bearing points from R-078's sketch, now baked into the code:

- **No `Retry-After` accessor exists on any SDK exception** (only `StatusCode` / `ResponseBody` / `ErrorType`). `ex.RetryAfter` would not compile. The spec's "retry-after from raw headers" is therefore honored via a `RetryAfterCaptureHandler` (`DelegatingHandler`) in the SDK's `HttpClient` pipeline — the only place the raw header is observable.
- **`AnthropicRateLimitException : Anthropic4xxException : AnthropicApiException`**, and 408/409 land on the non-leaf `Anthropic4xxException` — so catch order is specific-first and the final classification is by integer `StatusCode`, not leaf type.
- **`MaxRetries` default corrected `3 → 2`** (= 3 attempts, the SDK's own default, which the old value silently exceeded → 4 attempts). The stale `IPlanGenerationService` doc comment was fixed to match.

### Test evidence (full backend suite green: 1510/1510)

- `PlanAdaptationSchemaStabilityTests` — byte-stable schema, `additionalProperties:false`, required discriminators/slots, 20-keyword forbidden sweep.
- `PlanAdaptationOutputValidatorTests` — accept well-formed nudge/absorb/restructure; reject multi-slot, slot/kind mismatch, load-increase-under-non-Green, load-reducing-restructure-without-forward-path.
- `AdaptationPromptTests` — no `VDOT`, Daniels-Gilbert vocabulary, deviation/escalation tokens, spotlighting delimiters, and a load-through-the-prompt-store check (proves PR5's resolution path).
- `ClaudeCoachingLlmFailureTests` — end-to-end against a **real `AnthropicClient` + stub transport**: 429→Transient with retry-after read from raw headers, 400→Permanent, 503→Transient, network→Transient, `MaxRetries=2` ⇒ exactly 3 attempts, pre-cancelled token surfaces as `OperationCanceledException` (never reclassified), and the shared text path.
- `AdaptationResponseDtoTests` — `Kind=Error` envelope derivation (retryable + retry-after for transient; non-retryable for permanent).

### Deferred to PR5 (orchestration)

The trigger boundary (`EvaluateAdaptationCommand` + Wolverine handler), the actual restructure call wiring (rendering the prompt + invoking `GenerateStructuredAsync` + the one re-prompt on validator reject), the `PlanAdaptationOutput → PlanAdaptedFromLog`/diff mapping, the HTTP-200 controller that returns the `Kind=Error` envelope, and the success-path `AdaptationResponseKind.Adapted` payload. The live-recorded adaptation eval scenario is PR6.

### Follow-ups found

- **Flaky DEC-057 concurrency guard — `OnboardingTurnConcurrencyIntegrationTests` (pre-existing, not touched by this PR).** Intermittently fails with `successCount == 2` — two concurrent `StartStream<OnboardingView>` commits winning instead of one. Observed once in this PR's coverage-instrumented CI run; passed the normal build+test job, the re-run, and locally (1510/1510). Crucially, `found 2` is a **violation of the exactly-one-winner correctness property**, not a failure to reproduce the race — so it is either **(a)** a genuine rare gap in the Rich-mode (DEC-057) stream-collision guarantee (i.e. a real duplicate-onboarding-stream corruption risk) or **(b)** a test-harness contention artifact under coverage timing. Needs root-causing to distinguish; must **not** be masked with a retry/rerun. If (a) → fix the concurrency guarantee; if (b) → make the 5-way contention deterministic (e.g. a start barrier so all sessions reach `SaveChangesAsync` together). Will move to the cycle plan's "Captured During Cycle" table at slice completion.
- `AnthropicSchemaSanitizer` currently lives under the `Onboarding` namespace though it is general-purpose; a future move to `Models/Structured` would read cleaner (out of scope here).

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added plan adaptation capability enabling training plans to be adjusted through targeted nudges, structural changes, or absorption of modifications

* **Improvements**
  * Enhanced error handling with automatic retry logic for transient failures
  * Implemented server-guided retry timing for improved rate-limit management
  * Added comprehensive validation for adaptation responses to ensure consistency and safety


backend/tests/RunCoach.Api.Tests/Modules/Coaching/Adaptation/AdaptationPromptTests.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private static string GetAnchorBlock(string promptText, string anchor)
+    {
+        var anchorStart = promptText.IndexOf(anchor, StringComparison.Ordinal);
+        anchorStart.Should().BeGreaterThanOrEqualTo(
+            0,
+            because: $"the adaptation prompt must define a '{anchor}' block — without it, structural assertions can't proceed");
+        return promptText[anchorStart..];
+    }
```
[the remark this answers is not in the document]
leehopper: Addressed in 50946d6. Bounded `GetAnchorBlock` to the next top-level YAML key (or EOF) via a non-backtracking `\n(?=\S)` regex, so `context_template` token assertions can only match inside that block.

One note: the suggested `IndexOf("\n[a-zA-Z_]")` does a literal-substring search (not a pattern match), so it wouldn't have matched — used a regex instead. No false positive existed today since `context_template` is the file's last block, but this future-proofs against reordering.

leehopper (commented):
### Deep Review

Reviewed PR #167 (28 files, +1683/−16) across 7 dimensions (bugs, security, cross-file impact, tests, conventions/intent, type design, simplification) with validation + blind-challenge verification.

Found **11 findings** (0 critical, 1 high, 5 medium test-coverage gaps, 5 low): the high is a real failure-translation hole (SDK per-attempt timeout escapes the DEC-073 contract as a raw `TaskCanceledException`); the rest are an undocumented spec deviation (120s vs ~30s timeout), the post-response translation gap vs the new totality contract, test-coverage gaps on the new failure plumbing, and two doc/prompt consistency items.

**Notable verified non-issues:** the GATE-BEFORE-INCREASE self-reported-`NetLoadDelta` concern (flagged independently by 3 agents) was eliminated — the slice spec explicitly documents the one-validator design as intentional and nothing reaches the validator until PR5; the `AsyncLocal` array pattern, catch order vs the SDK exception hierarchy, prompt-injection delimiter setup, and the infinite `HttpClient.Timeout`-with-`ClientOptions.Timeout`-governor design all check out (the latter verified against the decompiled SDK).

Review methodology: 7 Fable review agents → 19 raw findings → deterministic verification (19 kept; orchestrator corrected a failed git-blame classification) → 5 independent validators (8 eliminated, mostly PR5-conditional) → rules filter → 11 blind challengers (all survived, 1 contested). 2j per-file summaries were folded into the change-summarizer dispatch.

---
Generated by deep-review | Reviewed up to: 50946d621ccccd7093d88c3d357e336b9a0963d5

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
```diff
@@ -368,4 +396,60 @@ private static void ValidateSettings(CoachingLlmSettings settings)
...
+                    ? new TransientCoachingLlmException(UnavailableMessage, RetryAfterCapture.CurrentSeconds, ex)
+                    : new PermanentCoachingLlmException(RejectedMessage, ex);
+            }
+            catch (AnthropicException ex)
+            {
+                // Defensive catch-all for any unmodeled SDK exception (e.g. a bare service error).
+                throw new PermanentCoachingLlmException(RejectedMessage, ex);
+            }
+        }
+    }
```
leehopper, line 503: **🟠 [HIGH] SDK per-attempt timeout escapes as raw TaskCanceledException — never translated to TransientCoachingLlmException**

The catch chain deliberately lets `OperationCanceledException` through, but the SDK's per-attempt timeout (`ClientOptions.Timeout`, 120s) fires as a `TaskCanceledException` from `HttpClient.SendAsync` via a linked CTS — with the caller's token **not** cancelled. SDK 12.24.1 has no timeout exception type (verified against assembly metadata: no `AnthropicTimeoutException`, no `TaskCanceledException` reference, no timeout wrapping — its own `ClientOptions.HttpClient` doc describes the conflict symptom as 'premature cancellation'), and its retry loop does not retry timeouts. Since the public ctor sets `HttpClient.Timeout = InfiniteTimeSpan`, the SDK timeout is the sole governor: a hung Anthropic call on the live onboarding/plan paths surfaces as a raw `TaskCanceledException`, bypassing the DEC-073 contract (`AdaptationResponseDto.FromError` only accepts `CoachingLlmException`) and getting misdiagnosed as user cancellation.

**Suggested fix:** add a filtered catch before the Anthropic catches:
```csharp
catch (OperationCanceledException ex) when (!ct.IsCancellationRequested)
{
    throw new TransientCoachingLlmException(UnavailableMessage, retryAfterSeconds: null, ex);
}
```
Genuine caller cancellation still propagates unwrapped. Consider a stub-transport test with a short `ClientOptions.Timeout` and a hanging handler.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs#L411-L455)

---
*Confidence: 85% | challenge: 75 | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/CoachingLlmSettings.cs
```diff
@@ -31,11 +31,12 @@ public sealed record CoachingLlmSettings
...
-    /// Gets maximum number of retries for failed requests (rate limits, transient errors).
-    /// The Anthropic SDK handles retries with exponential backoff.
-    /// Defaults to 3.
+    /// Gets the maximum number of retries the Anthropic SDK makes for failed requests (rate
+    /// limits, transient 5xx/network errors), with exponential backoff that honors
+    /// <c>Retry-After</c>. <c>MaxRetries = N</c> means up to <c>N + 1</c> attempts. Defaults to
+    /// 2 (3 attempts) per DEC-073, matching the SDK's own default.
     /// </summary>
-    public int MaxRetries { get; init; } = 3;
+    public int MaxRetries { get; init; } = 2;
```
leehopper, line 39: **💡 [LOW] DEC-073/spec require ~30s per-attempt timeout; default stays 120s with no deferral note**

DEC-073 (decision-log line ~2721: 'per-attempt Timeout ≈ 30s (down from the 10-min default)') and the slice-3 spec Unit 4 FR ('MaxRetries=2 → 3 attempts, ~30s/attempt') define both halves of the SDK config. This PR lands the MaxRetries half citing DEC-073 but leaves `TimeoutSeconds` at 120 (and the test still pins 120) with no deferral/supersession note in the spec, PR body, or comments. If 120s is the deliberate trade-off because live plan generation shares this setting, record that supersession (decision log or doc comment); otherwise consider a separate per-call timeout for the adaptation path in PR5.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/CoachingLlmSettings.cs#L33-L45)

---
*Confidence: 80% | challenge: 88 | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
```diff
@@ -186,7 +209,7 @@ public async Task<string> GenerateAsync(
             },
         };
 
-        var response = await _client.Messages.Create(createParams, ct).ConfigureAwait(false);
+        var response = await CreateMessageAsync(createParams, ct).ConfigureAwait(false);
```
leehopper, line 191: **💡 [LOW] DEC-073 translation misses post-response failures (max_tokens truncation, deserialize) — contradicts the totality contract this PR documents**

`CreateMessageAsync` wraps only the SDK call. The `stop_reason=max_tokens` truncation check (lines 226-231, and 152-157 on the text path) throws bare `InvalidOperationException`, and `JsonSerializer.Deserialize` (line 233) can throw `JsonException`/`InvalidOperationException` — all escaping `ICoachingLlm` untyped. That contradicts what this PR documents elsewhere: `CoachingLlmException` ('the only exceptions ICoachingLlm surfaces to callers'), the rewritten `IPlanGenerationService` failure semantics, and `AdaptationResponseDto.FromError(CoachingLlmException)`. PR5's envelope path would let a truncated restructure payload (realistic at MaxTokens) surface as a 5xx instead of `Kind=Error`. These lines pre-date the PR; it's the new totality contract that makes them load-bearing.

**Suggested fix:** translate truncation/deserialize failures into `PermanentCoachingLlmException` (DEC-073 classifies malformed-data as terminal), or scope the `CoachingLlmException`/`IPlanGenerationService` docs to SDK transport failures only.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs#L225-L236)

---
*Confidence: 78% | challenge: 85 | pre-existing lines, newly consequential | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
```diff
@@ -186,7 +209,7 @@ public async Task<string> GenerateAsync(
...
```
leehopper, line 192: **💡 [LOW] Same gap vs DEC-073's malformed-data-is-terminal classification (intent)**

Companion to the finding above, framed against the decision text: DEC-073 explicitly classifies malformed-data as **terminal** (should surface as `PermanentCoachingLlmException`), yet truncation (lines 226-231) and null-literal/malformed deserialization (233-235) escape as `InvalidOperationException`/`JsonException` and would reach `ErrorHandlingMiddleware` as a 5xx — contradicting `AdaptationResponseDto`'s documented 'never a 5xx' promise. Fixing the translation gap (previous comment) resolves both.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs#L225-L236)

---
*Confidence: 75% | challenge: 85 | pre-existing lines, newly consequential | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
```diff
@@ -368,4 +396,60 @@ private static void ValidateSettings(CoachingLlmSettings settings)
...
+            {
+                // Remaining HTTP errors: 408/409 land on the non-leaf Anthropic4xxException, so
+                // classify by status rather than leaf type. 408/409/429/5xx are transient (the
+                // SDK would have retried them); 400/401/403/404/422 are permanent.
+                var status = (int)ex.StatusCode;
+                var transient = status is 408 or 409 or 429 || status >= 500;
+                throw transient
+                    ? new TransientCoachingLlmException(UnavailableMessage, RetryAfterCapture.CurrentSeconds, ex)
+                    : new PermanentCoachingLlmException(RejectedMessage, ex);
+            }
```
leehopper, line 496: **🟡 [MEDIUM] Test gap: 408/409 transient classification is never exercised**

`status is 408 or 409 or 429 || status >= 500` — the transient side of this expression is unreachable by the current tests: 429 exits via `AnthropicRateLimitException`, 503 via `Anthropic5xxException`, transport via `AnthropicIOException`, and the 400 test only exercises the permanent side. 408/409 on the non-leaf `Anthropic4xxException` is the branch's stated reason for existing (per its comment) and is exactly the case the PR body calls out as a decompile-verified correction — worth a stub test sending 408 and 409 asserting `TransientCoachingLlmException`.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs#L436-L448)

---
*Confidence: 75% | challenge: 92 | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/Adaptation/PlanAdaptationOutputValidator.cs
```diff
@@ -0,0 +1,74 @@
...
+            AdaptationKind.Absorb => !hasNudge && !hasRestructure,
+            AdaptationKind.Nudge => hasNudge,
+            AdaptationKind.Restructure => hasRestructure,
+            _ => false,
+        };
+
+        if (!slotMatchesKind)
+        {
+            return PlanAdaptationOutputValidationResult.Invalid(
+                PlanAdaptationOutputValidationViolation.SlotKindMismatch);
```
leehopper, line 53: **🟡 [MEDIUM] Test gap: Nudge-arm-false and unknown-kind arms of the discriminator switch untested**

`PlanAdaptationOutputValidatorTests` falsifies only the Restructure arm (restructure-kind + nudge slot) and the Absorb arm (absorb-kind + slot). Missing: `AdaptationKind.Nudge` with `NudgePatch == null` (the Nudge arm never returns false — a mutation of that arm to `true` survives the suite), and the `_ => false` default arm for an out-of-range discriminator (`(AdaptationKind)99`; `JsonStringEnumConverter` accepts integer values by default, so this is deserializable). The empty-Nudge shape is a realistic LLM failure this validator exists to guard.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/Adaptation/PlanAdaptationOutputValidator.cs#L46-L54)

---
*Confidence: 70% | challenge: 90 | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/RetryAfterCapture.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    private static readonly AsyncLocal<int?[]?> Slot = new();
+
+    /// <summary>
+    /// Gets the retry-after value captured during the active scope, or <see langword="null"/>
+    /// when no scope is active or no <c>Retry-After</c> header was seen.
+    /// </summary>
+    public static int? CurrentSeconds => Slot.Value?[0];
+
+    /// <summary>
```
leehopper, line 27: **🟡 [MEDIUM] Test gap: AsyncLocal capture has no scope-isolation, staleness, or multi-attempt tests**

The parent-allocated single-element-array-through-AsyncLocal pattern is the concurrency-sensitive heart of the Retry-After plumbing, and none of its subtle properties are pinned: (a) two concurrent logical calls capture independently, (b) a disposed scope cannot leak a stale value into a later call on the same async context, (c) last attempt wins across SDK retries, (d) `Record` outside a scope is a no-op. The MaxRetries=2 test uses `Retry-After: 0` and asserts only `CallCount`. A small `RetryAfterCaptureTests` file (plus one multi-attempt test asserting the captured value changes across attempts) would lock the semantics the DEC-073 translation depends on.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/RetryAfterCapture.cs#L16-L28)

---
*Confidence: 67% | challenge: 92 | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/Adaptation/PlanAdaptationOutput.cs
```diff
@@ -0,0 +1,85 @@
...
+/// before/after diff deterministically, and appends the event.
+/// </para>
+/// <para>
+/// Numerical bounds (e.g. weekly-mileage-jump ceilings, the over-performance cap) live
+/// in the prompt and the property <see cref="DescriptionAttribute"/> text, NOT as
+/// <c>minimum</c>/<c>maximum</c> schema keywords (Anthropic rejects those with HTTP 400).
+/// </para>
+/// </remarks>
+public sealed record PlanAdaptationOutput
+{
```
leehopper, line 30: **🟡 [MEDIUM] Test gap: no test deserializes schema-conformant JSON into PlanAdaptationOutput**

The schema (snake_case, string enums, all 7 properties required) and the validator are never connected: no test feeds an adaptation-shaped JSON document (as constrained decoding would emit against `AdaptationSchema.Frozen`) through `StructuredOutputSerializerOptions` into `PlanAdaptationOutput`. Failure tests throw before deserialization; validator tests use object initializers; the only typed round-trip uses `MacroPlanOutput`, which exercises none of the adaptation-specific shapes: explicitly-null required slots (`nudge_patch`/`restructure_plan`), nullable `ReferralCategory`, cross-module enums. One success-path stub test (200 + valid payload) would close the schema↔deserializer loop before PR5 goes live.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/Adaptation/PlanAdaptationOutput.cs#L30-L32)

---
*Confidence: 63% | challenge: 90 | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
```diff
@@ -59,11 +70,22 @@ public ClaudeCoachingLlm(
...
+            Timeout = Timeout.InfiniteTimeSpan,
+        };
+
         _client = new AnthropicClient(new ClientOptions
         {
             ApiKey = _settings.ApiKey,
             MaxRetries = _settings.MaxRetries,
             Timeout = TimeSpan.FromSeconds(_settings.TimeoutSeconds),
+            HttpClient = _httpClient,
         });
```
leehopper: **🟡 [MEDIUM] Test gap: production HttpClient pipeline wiring is untested — tests hand-replicate it**

`ClaudeCoachingLlmFailureTests.BuildLlm` rebuilds this exact pipeline through the internal ctor (and has already drifted: it omits `ClientOptions.Timeout`), so drift between replica and the real constructor is invisible. The owned-`Dispose` branch (`_ownsClient=true` → `_httpClient?.Dispose()`) is never executed by any test. Partially inherent (the public ctor hardcodes `SocketsHttpHandler`, so behavioral coverage needs a seam), but worth either extracting the pipeline construction into an internal factory both paths share, or at least a Dispose test through the public ctor.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs#L77-L90)

---
*Confidence: 63% | challenge: 85 | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/Adaptation/AdaptationResponseDto.cs
```diff
@@ -0,0 +1,59 @@
...
+/// Flat response envelope for the synchronous adaptation flow (DEC-073). The log-create itself
+/// always commits, so a terminal coaching-LLM failure is surfaced as a <c>Kind=Error</c> envelope
+/// over HTTP 200 — never a 5xx — carrying flat <see cref="ErrorMessage"/> / <see cref="Retryable"/>
+/// / <see cref="RetryAfterSeconds"/> fields the frontend reads without any SDK dependency.
+/// </summary>
+/// <remarks>
+/// The success-path payload (the adaptation turn / plan re-render signal) is added by the
+/// orchestration layer in Slice 3 Unit 5; this slice ships the error shape and its derivation
+/// from a <see cref="CoachingLlmException"/>.
+/// </remarks>
```
leehopper: **💡 [LOW] REVIEW.md-banned forward references: 'Slice 3 Unit 5 will…' narration in new comments**

REVIEW.md § Code comments bans planning-phase forward references, naming the 'Unit 5 will' pattern explicitly. New comments here, in `AdaptationResponseKind.cs:5`, `PlanAdaptationOutput.cs:21`, and `AdaptationPromptTests.cs` narrate what the not-yet-built orchestration layer will add. Symbolic DEC/spec provenance is allowed; future-behavior narration is what reads stale after PR5 lands. Suggest rewording to present-tense scope (e.g. 'carries only the error shape; the success payload is outside this type's scope') or trimming to the DEC-073 reference.

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Modules/Coaching/Adaptation/AdaptationResponseDto.cs#L9-L14)

---
*Confidence: 70% | challenge: 90 | deep-review*

backend/src/RunCoach.Api/Prompts/adaptation.v1.yaml
```diff
@@ -0,0 +1,109 @@
...
+  - net_load_delta: the net change in weekly volume in kilometers (positive adds load, negative
+    reduces it). GATE-BEFORE-INCREASE: whenever safety_tier is not Green, net_load_delta MUST be
+    zero or negative — never increase volume or intensity under a flagged tier.
+  - restructure_plan.revised_weekly_targets: revised weekly volume (km) for upcoming meso weeks.
+    restructure_plan.revised_current_week_workouts: revised workouts for the current week only
+    (the only week with detailed daily workouts). restructure_plan.forward_path: REQUIRED whenever
+    you reduce load — state how and when volume ramps back, so the runner sees the path back, not
+    just the cut.
+  - referral_category: set Injury or RedS when the safety tier carries a professional referral
+    (Amber); otherwise None/null.
```
leehopper: **💡 [LOW] Prompt says 'otherwise None/null' for referral_category; schema description says null-only**

Two LLM-facing surfaces diverge: the `[Description]` on `PlanAdaptationOutput.ReferralCategory` (which the model sees in the schema) says 'Null when no referral applies', while this prompt line offers 'otherwise None/null'. `ReferralCategory.None` is a schema-legal string the model may emit, which deserializes to non-null `ReferralCategory.None` — so a PR5 consumer branching on `ReferralCategory is null` would treat a Green-tier 'None' as a live referral. Cheapest fix now: change this line to 'otherwise null'. Alternatively normalize None→null in the validator/orchestration. (Contested in blind challenge only because nothing consumes the field until PR5.)

[View in context](https://github.com/leehopper/personal-running-coach/blob/50946d621ccccd7093d88c3d357e336b9a0963d5/backend/src/RunCoach.Api/Prompts/adaptation.v1.yaml#L50-L53)

---
*Confidence: 60% | challenge: 70 | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
leehopper: Fixed in 010cf98. Added `catch (OperationCanceledException ex) when (!ct.IsCancellationRequested)` ahead of the Anthropic catches, translating the SDK per-attempt timeout to `TransientCoachingLlmException` (no retry-after). Genuine caller cancellation still propagates unwrapped — the existing pre-cancelled-token test pins that. Added the suggested stub test: hanging transport + `ClientOptions.Timeout` of 1s asserting `TransientCoachingLlmException`, which confirmed the raw `TaskCanceledException` escape before the fix.

backend/src/RunCoach.Api/Modules/Coaching/CoachingLlmSettings.cs
leehopper: Recorded the supersession in 010cf98 as a doc comment on `TimeoutSeconds`: 120s is deliberate because this single setting is shared with plan generation, whose structured-output responses routinely need far longer than the ~30s DEC-073 sketched for the adaptation call alone. A separate per-call adaptation timeout remains an option for the orchestration PR if the shared bound proves too loose.

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
leehopper: Fixed in 010cf98. `max_tokens` truncation (both text and structured paths, via a shared `ThrowIfTruncated`) and deserialize failures (`JsonException`, null-literal guard) now translate to `PermanentCoachingLlmException` with the original detail retained as the inner exception — so `CoachingLlmException` is now truly the only failure surface of `ICoachingLlm`. The four `ClaudeCoachingLlmTests` that pinned the old untyped contract were updated to assert the translated type + inner detail.

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
leehopper: Resolved by the same change (010cf98): malformed-data now surfaces as `PermanentCoachingLlmException` per DEC-073's terminal classification, so it reaches `AdaptationResponseDto.FromError` as a non-retryable envelope instead of escaping to `ErrorHandlingMiddleware` as a 5xx.

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
leehopper: Added in 010cf98: `GenerateStructuredAsync_TranslatesRetryable4xx_ToTransient`, a `[Theory]` sending 408 and 409 through the stub transport and asserting `TransientCoachingLlmException` — the status-based branch on the non-leaf `Anthropic4xxException` is now exercised on its transient side.

backend/src/RunCoach.Api/Modules/Coaching/Adaptation/PlanAdaptationOutputValidator.cs
leehopper: Added in 010cf98: `Validate_ReturnsSlotKindMismatch_WhenNudgeKindFillsNoSlot` (Nudge kind, both slots null — the empty-nudge LLM failure shape) and `Validate_ReturnsSlotKindMismatch_ForUnknownAdaptationKind` (`(AdaptationKind)99` falling to the `_ => false` default arm). Both arms of the discriminator switch are now falsified.

backend/src/RunCoach.Api/Modules/Coaching/RetryAfterCapture.cs
leehopper: Added in 010cf98: a new `RetryAfterCaptureTests` covering (a) concurrent logical calls capturing independently (two scopes recording after an await, each reading back only its own value), (b) a disposed scope not leaking a stale value into a later `Record`/read on the same async context, (c) last-write-wins across attempts within one scope, and (d) `Record` outside a scope as a no-op. Also added the end-to-end multi-attempt test in `ClaudeCoachingLlmFailureTests`: a per-attempt Retry-After schedule of [0, 0, 7] with MaxRetries=2 asserting the thrown exception carries 7 (the final attempt's header).

backend/src/RunCoach.Api/Modules/Coaching/Adaptation/PlanAdaptationOutput.cs
leehopper: Added in 010cf98: `GenerateStructuredAsync_DeserializesSchemaConformantAdaptationJson_OnSuccess` — a 200 stub response whose payload is hand-written exactly as constrained decoding emits against `AdaptationSchema.Frozen` (snake_case keys, string enums, explicit-null `nudge_patch`, nullable `referral_category`, a full `WorkoutOutput` with cross-module `WorkoutType`), flowing through `GenerateStructuredAsync` + `StructuredOutputSerializerOptions` into `PlanAdaptationOutput`, asserting the materialized values + usage, and finishing through the validator.

backend/src/RunCoach.Api/Modules/Coaching/ClaudeCoachingLlm.cs
leehopper: Fixed in 010cf98. Extracted `ClaudeCoachingLlm.CreateClientPipeline(settings, transportHandler)` as the single construction seam: the public ctor calls it with `SocketsHttpHandler`, and `BuildLlm` in the failure tests now calls it with the stub — which also picks up `ClientOptions.Timeout`, closing the exact drift flagged here. Added `Dispose_ThroughPublicConstructor_DisposesOwnedPipelineIdempotently` so the `_ownsClient=true` branch is executed.

backend/src/RunCoach.Api/Modules/Coaching/Adaptation/AdaptationResponseDto.cs
leehopper: Reworded in 010cf98: `AdaptationResponseDto` / `AdaptationResponseKind` / `PlanAdaptationOutput` docs now state present-tense scope ("owned by the orchestration layer", "outside this type's scope") and the prompt-test comments drop the "(Unit 5)" narration. DEC/spec provenance references kept.

backend/src/RunCoach.Api/Prompts/adaptation.v1.yaml
leehopper: Fixed in 010cf98 — the prompt line now reads "otherwise null", matching the schema `[Description]` ("Null when no referral applies"). Held off on a validator-side None→null normalization since nothing consumes the field yet; if live outputs show Green-tier `"None"` emissions despite the aligned guidance, normalization belongs in the orchestration layer.

PR #169, opened 2026-06-10
feat(slice-3): adaptation orchestration + trigger boundary (PR5)

leehopper:
## Summary

PR5 of the Slice 3 merge train — the single integration gate. Wires the full adaptation loop: a committed workout log now synchronously triggers `EvaluateAdaptationCommand` via `bus.InvokeForTenantAsync`, orchestrating SafetyGate → deviation/escalation → [LLM at L2 Green/Amber] → validators → Marten append, with `WorkoutLogId`-keyed co-transactional idempotency and Rich-AppendMode optimistic concurrency. Depends on PR1 (#161), PR3 (#163), PR2 (#165), PR4 (#167) — all merged.

## What shipped (by commit)

- `84d4c97` — `AdaptationSignalState` persistence resolved (deferred design question): small per-plan Marten document stored co-transactionally with event appends; validating `Create` factory at the rehydration boundary (clamps score, rejects negative streak, rejects `NeedsAdjustment` with null `LastAdaptationOn`) — PR2-review carryover #2.
- `f888344` — dead-zone re-escalation locked (PR2-review carryover #1): L1 key-miss nudges still fire inside the dead-zone, the 3-consecutive-missed-days hard trigger re-escalates once the 7-day cooldown elapses, suppressed triggers emit an observability log; policy version bumped to v1.1.0. Anti-flip-flop hysteresis for score-driven L2 unchanged.
- `0b34c2b` — `ComposeForAdaptationAsync` renders the versioned adaptation prompt with CSPRNG base64url nonce-spotlighted recent-log context; user-controlled text is sanitized and delimiter-escaped.
- `b704b11` — `EvaluateAdaptationCommand` + events-only Wolverine handler (deterministic paths): off-plan no-op, L0 absorb (no event), L1 deterministic swap (no LLM), Red short-circuit (scripted content routed by referral category — crisis vs emergency-referral copy), marker choreography (seen-first / record-last, nothing staged on failure).
- `67a1b67` — create-flow trigger boundary: unconditional post-commit dispatch from the create controller; `CreateWorkoutLogResponseDto` now carries the adaptation envelope as a required property (201 always — `Kind=Error` never fails the create); OpenAPI codegen regenerated.
- `bfabe75` — L2 restructure path: exactly one structured-output LLM call (frozen schema + cache control), deterministic `RestructureDiffCalculator` (projection-space diff, current-week-only micro edits), Amber dual-append, `Kind=Error` envelope with zero staged writes on terminal LLM failure.
- `2d3799a` — integration suite: all 13 Gherkin scenarios through the live HTTP + Wolverine pipeline; public scriptable `StubCoachingLlm` swapped in the app factory (zero real API calls; unscripted calls throw); deterministic barrier-based concurrency test (no retries, no masking).
- `d0bfd4a` — **bug found by the suite, fixed**: the Rich-mode append race throws `JasperFx.Events.EventStreamUnexpectedMaxEventIdException`, not `Marten.Exceptions.ConcurrentUpdateException` (verified by decompiling the pinned assemblies). Chain retry rule + controller catch retargeted; the concurrency loser now genuinely retries and replays the winner's idempotency marker.
- `6fd0ff0` — stale comments aligned with the retargeted exception type.

## Verification

- Full suite **1611/1611 green** (build 0 warnings/0 errors), four independent runs including the validator's; concurrency scenario stable across repeated isolated runs.
- 6-gate validation PASS: 6/6 requirements, 13/13 Gherkin scenarios, 9/9 proof artifacts; scope audit clean; credential scan clean.
- Frontend green against regenerated codegen (build + 458/458 tests).
- No real LLM calls anywhere in this PR — the live-call surfaces arrive in PR6 (eval cache record) and the post-slice live E2E pass.

## Follow-ups found

- [ ] **Global DEC-057 concurrency rules rest on the same disproved exception type** (`Program.cs`): a real Rich-mode append race on the onboarding/regenerate chains would match no global rule and fall into Wolverine's default retry loop, defeating first-write-wins. Plausible root-cause lead for the scheduled `OnboardingTurnConcurrencyIntegrationTests` two-winners flake investigation (#168) — deliberately untouched here; should be folded into that investigation.
- [ ] Adaptation idempotency markers expire after 48h (sweeper) while the create-side EF idempotency is permanent — a re-trigger >48h after the original adaptation would re-evaluate. Mostly theoretical under synchronous post-create dispatch; worth a guard decision before async dispatch arrives in Slice 4.

## Summary by CodeRabbit

* **New Features**
  * Workout plans now adapt automatically after you log a workout (deterministic micro-adjusts for minor misses; AI-powered restructures for sustained deviations).
  * Create-workout responses return an inline adaptation result (success or error) with guidance.

* **Safety**
  * New scripted emergency and amber referral messages provide clear, deterministic stop-and-refer guidance.

* **Bug Fixes / Reliability**
  * Improved idempotency, concurrency handling and error-mapping so retries and conflicts yield consistent, retryable responses.

* **Documentation**
  * Decision added clarifying MVP adaptation scope and prompt contents.


backend/src/RunCoach.Api/Modules/Training/Adaptation/AdaptationSignalState.cs
```diff
@@ -28,4 +31,87 @@ public sealed record AdaptationSignalState(
...
+            rollingDeviationScore,
+            0.0,
+            AdaptationThresholds.MaxRollingDeviationScore);
+
+        return new AdaptationSignalState(
+            planState,
+            clampedScore,
+            consecutiveMissedDays,
+            lastAdaptationOn);
+    }
```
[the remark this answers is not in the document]
leehopper: Fixed in 72fc1b1: `Create` now rejects undefined `PlanState` encodings with `ArgumentOutOfRangeException` before the other invariant checks, and `AdaptationSignalStateFactoryTests` covers -1, 3, and `int.MaxValue` encodings.

backend/tests/RunCoach.Api.Tests/Infrastructure/StubCoachingLlm.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        // The delegate may throw a CoachingLlmException to exercise the
+        // handler's terminal-failure envelope paths; anything it throws
+        // propagates exactly as the production adapter's failures would.
+        var outcome = behavior();
+        if (outcome is not T typed)
+        {
+            throw new InvalidOperationException(
+                $"StubCoachingLlm behavior produced {outcome.GetType().Name} but the caller "
+                + $"requested {typeof(T).Name}; script a matching structured output.");
```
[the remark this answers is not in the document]
leehopper: Fixed in c7257bf: the mismatch diagnostic now reports `null` via `outcome?.GetType().Name ?? "null"` instead of throwing `NullReferenceException` while building the message.

backend/openapi/swagger.json
```diff
@@ -1030,6 +1030,43 @@
...
+          "retryable"
+        ],
+        "type": "object",
+        "properties": {
+          "kind": {
+            "$ref": "#/components/schemas/AdaptationResponseKind"
+          },
+          "adaptationKind": {
+            "$ref": "#/components/schemas/AdaptationKind"
+          },
```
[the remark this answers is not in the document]
leehopper: Not fixing here — this is a known, repo-wide limitation with a deliberate deferral, not drift specific to this PR. `AdaptationKind` is C#-nullable on the DTO (`AdaptationKind?`); the `required`/non-null marking comes from `RequireNonNullablePropertiesSchemaFilter`, which promotes every `$ref` property regardless of C# nullability because Swashbuckle emits bare `$ref`s with no nullable wrapper for enum properties. The same pattern affects 40+ properties already shipping (e.g. `MesoDaySlotOutput.workoutType`). `swagger.json` is generated output behind the codegen drift gate, so a hand-edit would fail CI; the real fix is nullable-aware `$ref` emission in the schema filter, which was evaluated and deferred on PR #163. The schema is type-accuracy-only here — generated Zod schemas are not applied to responses at runtime, and consuming frontend modules hand-write accurate discriminated models for envelope types.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
```diff
@@ -0,0 +1,600 @@
...
+                        safety.Tier,
+                        BuildNudgeRationale(diff),
+                        diff));
+                session.Store(AdaptationSignalStateDocument.From(planId, decision.NextState));
+                LogNudgeApplied(logger, cmd.WorkoutLogId, planId, snapshot.WeekNumber);
+                return RecordAndReturn(idempotency, cmd.WorkoutLogId, AdaptationKind.Nudge);
+            }
+
+            LogMicroAdjustEscalated(logger, cmd.WorkoutLogId, planId, snapshot.WeekNumber);
+        }
```
leehopper, line 317: **🟡 [MEDIUM] Escalated L1->L2 restructure persists the MicroAdjust signal state, disabling the cooldown/dead-zone hysteresis**

When a MicroAdjust decision cannot be swapped (no live micro week or no valid forward swap), Handle falls into RestructureAsync, appends a PlanAdaptedFromLog with EscalationLevel.Restructure, but at line 451 stores decision.NextState — which is still the MicroAdjust state the classifier produced: PlanState.MinorDeviation (or the held dead-zone state) with LastAdaptationOn unchanged (null until a first direct restructure). The documented invariant in AdaptationSignalState.Create (a restructure is the only transition into NeedsAdjustment and it always stamps the date) is violated in the opposite direction: a restructure committed without the state transition. Consequence: after an escalated restructure the 7-day cooldown and score dead-zone never engage — the very next under-performing log pushes the rolling score from 2.0 to 3.0 and immediately fires a second LLM restructure, the exact flip-flop the DEC-012 asymmetric hysteresis exists to prevent (plus an unbudgeted extra LLM call per log). Secondary effect: the dead-zone L1 carve-out (missed key workout mid-cooldown, which the classifier deliberately answers with a Nudge to avoid re-firing) escalates into a full restructure DURING the active cooldown whenever the swap is unswappable, punching through the dead-zone the carve-out explicitly promises to hold. The in-PR unit tests for the escalated path (Handle_MicroAdjust_WithNoForwardSwap_EscalatesIntoTheLlmRestructure) assert the event level but never assert the stored signal-state document, so this is unpinned.

**Suggested fix:** On the escalated path, persist a restructure-shaped state instead of decision.NextState — e.g. AdaptationSignalState.Create(PlanState.NeedsAdjustment, decision.NextState.RollingDeviationScore, decision.NextState.ConsecutiveMissedDays, deviation.OccurredOn) — or have RestructureAsync derive the stored state from the executed level rather than the classified level. Add a unit test asserting the stored document after an escalated restructure enters NeedsAdjustment with LastAdaptationOn stamped.

---
*bug | confidence 100% | blind-challenge 90/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
```diff
@@ -0,0 +1,600 @@
...
+                        cmd.WorkoutLogId,
+                        AdaptationKind.Nudge,
+                        EscalationLevel.MicroAdjust,
+                        safety.Tier,
+                        BuildNudgeRationale(diff),
+                        diff));
+                session.Store(AdaptationSignalStateDocument.From(planId, decision.NextState));
+                LogNudgeApplied(logger, cmd.WorkoutLogId, planId, snapshot.WeekNumber);
+                return RecordAndReturn(idempotency, cmd.WorkoutLogId, AdaptationKind.Nudge);
+            }
```
leehopper, line 316: **🟡 [MEDIUM] Score-driven L1 swaps a COMPLETED key workout and emits a false "You missed" rationale**

The handler attempts MicroAdjustPlanner.TryPlanSwap for EVERY MicroAdjust decision, but the classifier fires L1 on two distinct triggers: a skipped key workout, OR the rolling deviation score reaching MicroAdjustEnterScore (2.0) from accumulated under-performance — and an under-performing log can be COMPLETED (slow pace or short distance). Reachable sequence: log 1 is a slow easy run (score 1.0, absorb), log 2 is a completed-but-slow Tuesday tempo (score 2.0, L1). The handler then calls TryPlanSwap with snapshot.DayOfWeek of the tempo the runner JUST COMPLETED; TryPlanSwap only checks that the day holds a key workout (never the completion status), finds a valid forward swap, and the handler appends a Nudge that reschedules the already-done tempo to later in the week with BuildNudgeRationale rendering "You missed {Title} on Tuesday, so I moved it forward..." — a factually false user-facing message and a wrong plan mutation (the runner is told to redo a workout they completed). Conversely, when the score-driven L1 lands on a non-key day, TryPlanSwap returns false and the handler upgrades a score-2.0 signal to a full LLM restructure even though the restructure enter threshold is 3.0 — the over-escalation route bypasses the ladder thresholds entirely. The EscalationDecision record carries no trigger reason, so the handler currently cannot distinguish the two L1 causes.

**Suggested fix:** Gate the swap on the actual miss: attempt TryPlanSwap only when deviation.CompletionStatus == Skipped && deviation.IsKeyWorkout (e.g. carry the trigger reason on EscalationDecision or re-check the deviation in the handler). Decide explicitly what a score-driven L1 should do — a deterministic no-event nudge/absorb is closer to the DEC-012 ladder than silently escalating a 2.0-score signal into an LLM restructure reserved for 3.0.

---
*bug | confidence 95% | blind-challenge 90/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
```diff
@@ -0,0 +1,600 @@
...
+        // level; the LLM never absorbs a sustained deviation away) — maps to the
+        // same non-retryable Kind=Error envelope with nothing staged.
+        var validation = PlanAdaptationOutputValidator.Validate(output);
+        if (!validation.IsValid || output.AdaptationKind != AdaptationKind.Restructure)
+        {
+            LogRestructureProposalRejected(
+                logger, cmd.WorkoutLogId, planId, validation.Violation, output.AdaptationKind);
+            return AdaptationResponseDto.FromError(
+                new PermanentCoachingLlmException(RejectedProposalMessage, null));
+        }
```
leehopper, line 473: **🟡 [MEDIUM] GATE-BEFORE-INCREASE is enforced against the LLM-echoed SafetyTier, not the deterministic gate tier**

The handler validates the L2 proposal via PlanAdaptationOutputValidator.Validate(output) and stages the restructure when it passes, but the GATE-BEFORE-INCREASE check inside the validator (PlanAdaptationOutputValidator.cs line 57) compares output.NetLoadDelta against output.SafetyTier — the tier the LLM echoed back — and the handler never asserts that output.SafetyTier equals safety.Tier (the deterministic gate result it holds in scope). If the LLM echoes Green while the gate classified Amber, a proposal with NetLoadDelta above zero passes validation and a load-increasing restructure commits under a non-Green tier (the event even records safety.Tier=Amber and appends the Amber referral alongside the increase). The spec premise is that this invariant must never depend on the LLM policing itself; the comment at lines 441-442 documents exactly the trust gap. The prompt instructs echo-back, but instruction is not enforcement.

**Suggested fix:** In the post-decode policy block, reject the proposal (same non-retryable envelope) when output.SafetyTier differs from safety.Tier, or pass the deterministic tier into the validator so the NetLoadDelta clamp keys off the gate classification instead of the echo. Add a unit test: Amber gate + echoed-Green output + positive NetLoadDelta must be rejected with nothing staged.

---
*intent | confidence 85% | blind-challenge 92/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogsController.cs
```diff
@@ -162,6 +209,49 @@ private static partial void LogCreateRejectedInvalid(
...
+            // different, unrelated hierarchy (session-level write collisions) and
+            // never fires for this conflict.
+            LogAdaptationDispatchConflicted(logger, workoutLogId, userId, ex);
+            return new AdaptationResponseDto
+            {
+                Kind = AdaptationResponseKind.Error,
+                ErrorMessage = "Your workout was saved, but your coach could not review it just now. Submitting the same log again will retry the review.",
+                Retryable = true,
+            };
+        }
```
leehopper, line 269: **🟡 [MEDIUM] Never-fail-the-create contract untested for the other two recognized concurrency surfaces (ConcurrentUpdateException, DocumentAlreadyExistsException)**

EvaluateAdaptationAsync catches only EventStreamUnexpectedMaxEventIdException and maps it to a retryable Kind=Error envelope, with the XML doc asserting an adaptation failure never fails the create. But the integration suite added in this PR (AdaptationOrchestrationIntegrationTests.InvokeEvaluationAsync, lines 908-927) recognizes two additional legal loser surfaces of the same race: ConcurrentUpdateException (signal-state document collision) and DocumentAlreadyExistsException (the duplicate WorkoutLogId-keyed marker Insert — MartenIdempotencyStore.Record uses session.Insert at line 80, so this is reachable). Neither is retried by the chain rule (keyed solely on EventStreamUnexpectedMaxEventIdException) nor caught by the controller, so through the HTTP create path they would surface as a 5xx AFTER the log row committed. The concurrency integration test deliberately drives the race through the bus, not the controller, so no test covers the create endpoint under these surfaces and no test would fail when they break the documented contract.

_Challenge nuance: the ConcurrentUpdateException arm is probably NOT reachable in production — AdaptationSignalStateDocument has no optimistic concurrency, so its Store is last-write-wins; DocumentAlreadyExistsException via the marker session.Insert IS reachable (concurrent duplicate creates, absorb-path loser stages only the marker)._

**Suggested fix:** Add controller unit tests stubbing the bus to throw ConcurrentUpdateException and DocumentAlreadyExistsException and decide the contract: either widen the catch (and assert 201 + retryable error envelope, since a loser rollback means the winning adaptation committed) or pin the current 5xx as deliberate. Optionally add an HTTP-path variant of the concurrency race asserting both submitters receive 2xx.

---
*test_coverage | confidence 78% | blind-challenge 82/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
```diff
@@ -0,0 +1,600 @@
...
+        // protocol violation; throwing aborts the transaction with nothing staged.
+        if (plan is null)
+        {
+            throw new InvalidOperationException(
+                $"Cannot restructure: no plan projection found for plan {planId}.");
+        }
+
+        var composition = await assembler
+            .ComposeForAdaptationAsync(plan, decision.EscalationLevel, safety.Tier, deviation, detail, ct)
+            .ConfigureAwait(false);
```
leehopper, line 433: **💡 [LOW] Sanitization state is invisible in LoggedWorkoutDetail, and the handler double-sanitizes the triggering log before ComposeForAdaptationAsync sanitizes it again**

Encapsulation: 6/10, Expression: 3/10, Usefulness: 8/10, Enforcement: 3/10. LoggedWorkoutDetail carries no type-level distinction between raw and sanitized text, and the three IContextAssembler compose methods encode three conflicting doc-only sanitization contracts: ComposeForPlanGenerationAsync says the caller MUST pre-sanitize, ComposeForAdaptationAsync says it sanitizes internally, and EvaluateAdaptationHandler.SanitizeDetailAsync sanitizes the detail for the safety gate then reuses the SAME sanitized instance for the L2 prompt. Because LayeredPromptSanitizer.SanitizeAsync is not idempotent — Tier 3 unconditionally wraps the body in <WORKOUT_NOTE>...</WORKOUT_NOTE> and EscapeDelimiterBody escapes &/</> — the note reaching the LLM is wrapped twice and escaped three times (handler sanitize, assembler sanitize at ContextAssembler.cs line 377, spotlight escape at line 378), turning a benign & into &amp;amp;amp; and nesting literal delimiter tags. The assembler comment at lines 370-376 only accounts for its own two internal passes as intentional defense-in-depth; the handler-side third pass is unaccounted for. ContextAssemblerAdaptationTests prove the assembler expects RAW input (the fake-sanitizer sentinel test asserts the raw note must not appear), and no integration test asserts prompt content, so the garbling ships unobserved. Tier-2 audit findings (EventId 4001) also fire twice per note for the same content, double-counting injection telemetry.

**Suggested fix:** Make sanitization state structural: have IRecentLogSanitizer return a distinct SanitizedLoggedWorkoutDetail (or wrap Notes/Metrics in a SanitizedText record) that ComposeForAdaptationAsync does NOT re-sanitize, and have the safety gate accept the same sanitized type. Alternatively keep ComposeForAdaptationAsync as the single sanitization owner and pass the raw detail there while the handler sanitizes a separate copy for the gate — but encode whichever contract wins in the parameter type, not XML docs.

---
*type_design | confidence 95% | blind-challenge 92/100 | origin: surfaced | deep-review*

leehopper, line 433: **💡 [LOW] Escalated L1->L2 prompt renders "Escalation level: MicroAdjust" while the validator hard-requires a Restructure proposal**

RestructureAsync composes the adaptation prompt with decision.EscalationLevel, which on the unswappable-L1 escalation path is still EscalationLevel.MicroAdjust. The rendered context block therefore says "Escalation level: MicroAdjust" while the system prompt (adaptation.v1.yaml) asserts "a deterministic system... decided that a restructure is warranted" and instructs adaptation_kind is "normally Restructure for this call". Line 416 then rejects ANY proposal whose AdaptationKind != Restructure, mapping it to a permanent user-facing error ("did not pass safety validation") with nothing staged. An LLM that aligns its adaptation_kind with the echoed escalation level — the more faithful reading of the contradictory context — returns Nudge and the runner gets a terminal error envelope for a perfectly evaluable log; even when the LLM complies, the prompt carries contradictory deterministic facts on a path whose whole design is verbatim echo-back of deterministic values. The event appended on this same path already records the executed level (EscalationLevel.Restructure, line 434), so the prompt is the only place the stale classified level leaks.

**Suggested fix:** Pass the executed level — EscalationLevel.Restructure — to ComposeForAdaptationAsync on the escalated path (mirror the event), or thread an explicit executedLevel variable through RestructureAsync so prompt, event, and validator policy all agree.

---
*bug | confidence 90% | blind-challenge 90/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/ContextAssembler.cs
```diff
@@ -148,13 +164,21 @@ public sealed partial class ContextAssembler : IContextAssembler
...
+    /// </param>
     public ContextAssembler(
         IPromptStore promptStore,
         TimeProvider timeProvider,
         IPromptSanitizer sanitizer,
         IHostEnvironment environment,
         IOptions<PromptStoreSettings> promptSettings,
-        ILogger<ContextAssembler> logger)
+        ILogger<ContextAssembler> logger,
+        IRecentLogSanitizer? recentLogSanitizer = null)
```
leehopper, line 181: **💡 [LOW] ContextAssembler public constructor grew to 7 parameters but the load-bearing 6-arg DI comments and guard-test docs were not updated**

This PR adds a trailing optional IRecentLogSanitizer parameter to the ContextAssembler public constructor (now 7 parameters). Two load-bearing comments elsewhere still describe a single 6-arg public constructor: the DI registration comment in ServiceCollectionExtensions.cs (lines 63-74), which documents WHY the plain type registration is unambiguous and must not become a lambda factory (Wolverine codegen invariant, DEC-071), and the XML doc plus assertion reason in ContextAssemblerDiResolutionTests.cs (lines 11-12 and 58), which guard against the constructor-selection production bug. These comments are the canonical record of a real prior incident; an engineer relying on them to evaluate constructor-selection risk now reads an inaccurate constructor surface. The author demonstrated comment-alignment discipline elsewhere in this branch (commit 6fd0ff0 aligned stale concurrency comments) but missed these two files.

**Suggested fix:** Update both comment sites to describe the 7-arg constructor (6 required + optional IRecentLogSanitizer) and note that the optional parameter resolves from the registered singleton in production, keeping the DI-selection rationale accurate.

---
*cross_file_impact | confidence 85% | blind-challenge 92/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/ContextAssembler.cs
```diff
@@ -311,6 +337,67 @@ public async Task<PlanGenerationPromptComposition> ComposeForPlanGenerationAsync
...
+        // re-classifies them. The nonce is fresh per call (R-068 / DEC-059).
+        var tokens = new Dictionary<string, string>(StringComparer.Ordinal)
+        {
+            ["plan_context"] = BuildAdaptationPlanContext(plan),
+            ["escalation_level"] = escalationLevel.ToString(),
+            ["safety_tier"] = safetyTier.ToString(),
+            ["deviation_summary"] = BuildDeviationSummary(deviation),
+            ["recent_logs"] = recentLogs,
+            ["recent_logs_nonce"] = LayeredPromptSanitizer.GenerateSpotlightNonce(),
+        };
```
leehopper, line 396: **💡 [LOW] Adaptation prompt context omits the runner profile the spec lists as a prompt input**

Spec Unit 4 says the adaptation prompt consumes the ContextAssembler context of profile + plan projection + recent logs including notes/metrics. The shipped ComposeForAdaptationAsync renders plan context, escalation level, safety tier, deviation summary, and the triggering log only — no runner-profile block, and the adaptation.v1.yaml template defines no profile token. The recent-logs-window omission is acknowledged in the new XML doc as deferred, but the profile omission is acknowledged nowhere (code, spec, or decision log). Practical effect: the L2 restructure LLM proposes plan changes without the runner goal/background context the spec intended it to have. Confidence is moderated because the template (which fixes the token set) landed in the earlier PR4 and this PR only fills it.

**Suggested fix:** Either add the profile block to the adaptation composition (and template, with an eval-cache re-record), or document the omission as deliberate MVP-0 scope in the spec / a decision-log note so the intent record matches the implementation.

---
*intent | confidence 72% | blind-challenge 90/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper, line 473: **💡 [LOW] Validator-rejected proposal terminally fails immediately — the spec and design require one corrective re-prompt first**

Both intent documents specify a single re-prompt with constraint feedback when the post-decode validator rejects an LLM proposal, before declaring terminal failure. The handler instead makes exactly one structured-output call and maps any validation reject (or non-restructure proposal) straight to the non-retryable Kind=Error envelope; the unit test Handle_ValidatorRejectedOutput_ReturnsNonRetryableErrorEnvelopeWithNothingStaged pins "No second LLM attempt". The handler XML doc justifies single-call as "DEC-073 — SDK-side retries only", but DEC-073 covers transport-level SDK retries (429/5xx), not validator-reject re-prompts; neither DEC-073 nor DEC-079 supersedes the re-prompt requirement, and no decision record documents dropping it. The behavioral consequence: a runner whose sustained deviation crossed the L2 threshold gets an error envelope and no adaptation on a recoverable reject that one corrective re-prompt might have fixed.

**Suggested fix:** Either implement the spec-mandated single re-prompt with the validator violation fed back as constraint feedback (then terminal), or record the deviation explicitly — update the spec/design and add a decision-log note stating that validator rejects are terminal with zero re-prompts and why — so spec and implementation agree.

---
*intent | confidence 70% | blind-challenge 88/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogsController.cs
```diff
@@ -66,25 +97,35 @@ public async Task<IActionResult> CreateLog(
...
+        // Dispatch unconditionally — also on a DEC-077 replay that returned an
+        // EXISTING log id. The handler's WorkoutLogId-keyed idempotency marker
+        // makes a re-dispatch for an already-evaluated log a designed no-op,
+        // and a marker miss (crash before the Marten commit) lets the replayed
+        // create recover the missing evaluation.
+        var adaptation = await EvaluateAdaptationAsync(workoutLogId, userId, ct);
+
+        return Created(
+            $"/api/v1/workouts/logs/{workoutLogId}",
+            new CreateWorkoutLogResponseDto(workoutLogId, adaptation));
```
leehopper, line 133: **💡 [LOW] Log create now mutates the plan read model, but the frontend create mutation never invalidates the Plan cache tag**

After this PR, POST /api/v1/workouts/logs can change the plan a user is looking at: an L1 nudge swaps workouts and an L2 restructure edits weekly targets via inline PlanProjection updates. The shipped frontend create mutation (workout-log.api.ts createWorkoutLog) invalidates only the WorkoutLog tag, while getCurrentPlan caches under the Plan tag — so after logging a workout that triggers an adaptation, the home plan surface keeps rendering the stale pre-adaptation plan until an unrelated refetch. The new AdaptationResponseDto XML doc asserts the frontend refetches by invalidating its query tags after every successful log create, but that invalidation does not exist in any shipped frontend code — the only consumer update in this PR is generated codegen. If this is deliberately sequenced into Slice 3 Unit 7, it should be explicitly tracked; otherwise it is a one-line fix now.

_Confirmed deliberate sequencing: pr-strategy.md assigns Plan/Conversation tag invalidation to PR7 of this merge train. Residual concerns: the tracking lives only in a gitignored working-tree file, and the shipped AdaptationResponseDto XML doc asserts invalidation behavior that does not exist until PR7 lands._

**Suggested fix:** Add Plan (and the future conversation tag) to createWorkoutLog invalidatesTags — optionally only when response.adaptation.adaptationKind is Nudge or Restructure via the invalidatesTags callback — or explicitly confirm the Unit 7 PR in the merge train carries this wiring before deploy.

---
*cross_file_impact | confidence 30% | blind-challenge 85/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
```diff
@@ -0,0 +1,600 @@
...
+            usage.CacheReadInputTokens);
+        return RecordAndReturn(idempotency, cmd.WorkoutLogId, AdaptationKind.Restructure);
+    }
+
+    /// <summary>
+    /// Records the no-plan-change response under the log's id and returns it —
+    /// the shared tail of every committing deterministic path. The marker is
+    /// always the LAST thing staged so it commits atomically with whatever the
+    /// path appended before it.
+    /// </summary>
```
leehopper, line 550: **💡 [LOW] RecordAndReturn XML doc says it records the no-plan-change response on deterministic paths only, but it is also the tail of the nudge and LLM restructure paths**

The summary states that this method records the no-plan-change response and is the shared tail of every committing deterministic path. Both claims are inaccurate: the method is called with AdaptationKind.Nudge (a committed plan change) and from RestructureAsync with AdaptationKind.Restructure at line 460 — a plan change on the LLM path, which is neither deterministic nor no-plan-change. Only the off-plan, Red, and L0 absorb callers match the description. A reader auditing the marker choreography from this doc would wrongly conclude the restructure path records its marker elsewhere.

**Suggested fix:** Reword the summary to: Records the Kind=Adapted response for the resolved plan-change kind and returns it — the shared tail of every committing path; the marker is always the LAST thing staged.

---
*comment_accuracy | confidence 75% | blind-challenge 90/100 | origin: new | deep-review*

backend/tests/RunCoach.Api.Tests/Infrastructure/StubCoachingLlm.cs
```diff
@@ -0,0 +1,123 @@
...
+/// and the integration tier makes ZERO real API calls by contract — eval-cached
+/// LLM coverage lives in the eval tier, not here.
+/// </summary>
+/// <remarks>
+/// <para>
+/// The class must be <c>public</c>: Wolverine's codegen service-locates internal
+/// types and silently falls back to scope-based DI for the whole chain, which
+/// resolves a different <c>IDocumentSession</c> for the idempotency store than
+/// the one Wolverine commits and breaks same-session idempotency (see
+/// <c>backend/CLAUDE.md</c> § Wolverine Handler Integration Coverage).
```
leehopper: **💡 [LOW] New comment references an external doc path (backend/CLAUDE.md section) — comments must be self-contained**

The class remarks explain the public-visibility requirement by pointing the reader at another file: see backend/CLAUDE.md section Wolverine Handler Integration Coverage. The root REVIEW.md code-comments rules prohibit embedded repo file paths and external doc references in comments; only Spec / DEC / R-NNN symbolic references are allowed. The surrounding sentence already carries the load-bearing why (Wolverine codegen service-locates internal types, falls back to scope-based DI, and breaks same-session idempotency), so the pointer can simply be dropped.

**Suggested fix:** Delete the parenthetical doc-path reference; the in-comment explanation already stands alone.

---
*convention | confidence 75% | blind-challenge 92/100 | origin: new | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/WorkoutMetricsProjection.cs
```diff
@@ -0,0 +1,65 @@
...
+        if (string.IsNullOrEmpty(metricsJson))
+        {
+            return Empty;
+        }
+
+        var raw = JsonSerializer.Deserialize<Dictionary<string, JsonElement>>(metricsJson);
+        if (raw is null || raw.Count == 0)
+        {
+            return Empty;
+        }
```
leehopper, line 45: **💡 [LOW] Documented fail-loudly JsonException contract for malformed stored metrics is untested**

ToDisplayMetrics documents a deliberate safety-relevant contract: a stored Metrics string that is not a JSON object is data corruption and MUST throw JsonException loudly rather than silently skipping the safety-gate scan of its free-text values (the projected dictionary feeds ISafetyGate.Classify in the handler at EvaluateAdaptationHandler.SanitizeDetailAsync). The in-PR tests cover only the null/empty short-circuit and the happy-path scalar mapping (EvaluateAdaptationHandlerUnitTests lines 673-706); no test feeds malformed JSON (e.g. a bare array, truncated object, or non-JSON text) and asserts the throw. The fail-loudly behavior is exactly the kind of contract a well-meaning future change would weaken by wrapping the Deserialize call in a try/catch returning Empty — and nothing would fail.

**Suggested fix:** Add a Theory feeding malformed inputs — not-JSON text, a JSON array like [1,2], a truncated object — and assert Action act = () => WorkoutMetricsProjection.ToDisplayMetrics(input); act.Should().Throw<JsonException>().

---
*test_coverage | confidence 70% | blind-challenge 92/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Coaching/IContextAssembler.cs
```diff
@@ -83,6 +87,54 @@ Task<PlanGenerationPromptComposition> ComposeForPlanGenerationAsync(
...
+    /// The deterministic deviation measurement; its numbers are rendered
+    /// verbatim so the LLM performs no deviation math.
+    /// </param>
+    /// <param name="triggeringLog">
+    /// The logged workout that triggered the adaptation. Its note and
+    /// free-text metric values are sanitized inside the assembler via
+    /// <c>IRecentLogSanitizer</c> (R-068 / DEC-059), and the rendered line is
+    /// delimiter-escaped before being placed in the nonce-delimited
+    /// recent-logs section. The full recent-logs window is a later unit —
+    /// this composes the triggering log plus plan-week context only.
```
leehopper: **💡 [LOW] New XML doc carries a planning-phase forward reference: the full recent-logs window is a later unit**

The triggeringLog parameter doc on ComposeForAdaptationAsync ends with: The full recent-logs window is a later unit — this composes the triggering log plus plan-week context only. REVIEW.md requires comments to be productionized before merge with no planning-phase forward references (the rule explicitly lists the Unit 5 will / later slices will pattern). The deferral itself is real (spec Unit 4 expected the prompt context to include recent logs and the runner profile; the shipped composition renders only the triggering log + plan context), but the comment should state current behavior without pointing at future work.

**Suggested fix:** Rephrase as a statement of current scope, e.g.: At MVP-0 this composes the triggering log plus plan-week context only; the recent-logs section carries a single log.

---
*convention | confidence 70% | blind-challenge 90/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/EscalationClassifier.cs
```diff
@@ -7,11 +7,17 @@ namespace RunCoach.Api.Modules.Training.Adaptation;
...
 /// resolves L0 absorb / L1 micro-adjust / L2 restructure from a
 /// <see cref="DeviationResult"/>, the safety tier, and a prior signal state, applying
-/// asymmetric enter/exit hysteresis so a restructure cannot flip-flop. Pure and
-/// stateless — no LLM, no I/O; the L2 restructure is where PR5 hands off to the LLM.
+/// asymmetric enter/exit hysteresis so a restructure cannot flip-flop. The
+/// anti-flip-flop guarantee is L2-only: inside the dead-zone a missed key workout
+/// still earns an L1 micro-adjust, and a fresh consecutive-missed-days hard trigger
+/// re-escalates once the cooldown has elapsed. Pure and stateless — no LLM, no I/O
+/// beyond suppressed-trigger observability logging; the L2 restructure is where PR5
+/// hands off to the LLM.
```
leehopper: **💡 [LOW] Rewritten class doc retains the forward reference: the L2 restructure is where PR5 hands off to the LLM**

This PR rewrites the EscalationClassifier class summary (adding the dead-zone carve-out description) and keeps the phrase: the L2 restructure is where PR5 hands off to the LLM. PR5 is this very PR — after merge the reference is stale planning narrative rather than a symbolic spec/DEC reference, and the handoff target now has a name (EvaluateAdaptationHandler). REVIEW.md prohibits planning-phase forward references in productionized comments; PR numbers are not in the allowed Spec / DEC / R-NNN set. Note the same phrase also exists at line 99 in an unmodified body comment (pre-existing, listed here only for context).

**Suggested fix:** Replace with a code-anchored statement, e.g.: the L2 restructure is where EvaluateAdaptationHandler hands off to the coaching LLM.

---
*convention | confidence 67% | blind-challenge 85/100 | origin: new | deep-review*

leehopper (commented):
### Deep Review — PR #169 (adaptation orchestration + trigger boundary)

Found **19 findings** (0 critical, 0 high, **4 medium**, 7 low, 8 improvement suggestions). All 19 survived a blind-challenge verification round (scores 75-92/100). 18 are posted as inline comments; one finding targets a file outside this PR's diff and is included below.

**Medium (inline):** escalated L1->L2 restructure persists the MicroAdjust signal state (hysteresis never engages); score-driven L1 swaps a COMPLETED key workout with a false "You missed" rationale; GATE-BEFORE-INCREASE keys off the LLM-echoed SafetyTier; controller catch-scope gap lets `DocumentAlreadyExistsException` 5xx the create after commit.

#### 💡 [low] [security] Prompt-injection sanitizer bypassed by string values in numeric metric keys reaching the adaptation LLM prompt

This finding references `backend/src/RunCoach.Api/Modules/Coaching/Sanitization/RecentLogSanitizer.cs` (lines 15-63), which is not in this PR's diff.

The new L2 adaptation path feeds the triggering log into an LLM prompt. The metrics bag is persisted verbatim at create time (WorkoutLogService.SerializeMetrics) with no per-key type validation or key allowlist, so a client can submit a string value under a numeric metric key, e.g. hrAvg set to adversarial instruction text. WorkoutMetricsProjection.ToDisplayMetrics passes through every JSON string value verbatim regardless of key, and RecentLogFormatter renders any key present in WorkoutMetricKeys.Metadata such as hrAvg, power, cadence. But RecentLogSanitizer only routes the Weather and Terrain values through the DEC-059 IPromptSanitizer pattern catalog (FreeTextMetricKeys). A string stuffed into hrAvg therefore renders into the nonce-spotlighted recent-logs section of the adaptation prompt having bypassed the pattern sanitizer entirely; only LayeredPromptSanitizer.EscapeDelimiterBody applies. Delimiter escaping plus the spotlight nonce still prevent the attacker from closing the section, so this is a defense-in-depth gap rather than a full break-out, but it contradicts the documented coverage (the assembler XML doc claims it sanitizes the triggering log note and free-text metric values) and lets arbitrary instruction text reach the coaching model inside the untrusted block.

**Suggested fix:** Validate the metrics bag at the create boundary so numeric canonical keys reject non-numeric JSON values and reject keys outside WorkoutMetricKeys.All, OR route every rendered metric value through the DEC-059 IPromptSanitizer rather than only Weather/Terrain. Either closes the gap so no metric value reaches the LLM prompt without pattern sanitization.

*security | confidence 80% | blind-challenge 76/100 | origin: surfaced*

<details>
<summary>Review methodology</summary>

- Mode: all agents on Fable (user override of REVIEW.md frontier tier). 7 review dimensions dispatched (security-reviewer retried once after an API socket error; second run completed).
- Pipeline: 24 raw findings -> 24 verified (5 new / 19 surfaced, 0 eliminated) -> validated by 6 independent agents -> 17 passed Phase 6 filters (3 below threshold, 2 true duplicates deduped) -> +2 orchestrator reinstatements -> 19 blind-challenged, 19 survived (0 removed / 0 downgraded).
- Orchestrator overrides (documented): `test-1` reinstated after the injection filter false-positived on the 40-char identifier `EventStreamUnexpectedMaxEventIdException` matching its base64-payload regex; `type-1` reinstated after line-range dedup conflated it with the unrelated `bug-3` at the same call site (both independently challenge-confirmed at 90+).
- Author-acknowledged follow-ups (not re-reported): stale global DEC-057 exception rules in Program.cs; 48h adaptation marker expiry vs permanent EF create idempotency.

</details>

---
Generated by deep-review | Reviewed up to: c7257bf6a95048e08bc3cdd20e9dc78ad54cc2b0

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
```diff
@@ -0,0 +1,600 @@
...
+public sealed partial class EvaluateAdaptationHandler
+{
+    /// <summary>
+    /// Bounded inline retries for a lost stream-version race before the message
+    /// dead-letters. Each retry re-runs <see cref="Handle"/> in full: the marker
+    /// rolled back with the failed transaction so <c>SeenAsync</c> misses, and the
+    /// signal state reloads fresh — which IS the retry-against-fresh-state
+    /// semantics the optimistic-concurrency contract asks for.
+    /// </summary>
+    internal const int MaxConcurrencyRetries = 3;
```
leehopper, line 94: **💡 [LOW] Comments claim the exhausted concurrency retry dead-letters, but this chain is only invoked inline where MoveToErrorQueue is inert**

The MaxConcurrencyRetries doc (line 82-84: bounded inline retries for a lost stream-version race before the message dead-letters) and the Configure doc (line 116: retries inline a bounded number of times, then dead-letters) describe a dead-letter outcome. EvaluateAdaptationCommand is only ever dispatched via bus.InvokeForTenantAsync (controller + tests); per the DEC-073 verified note, Wolverine applies only Retry/RetryWithCooldown to inline InvokeAsync/InvokeForTenantAsync — MoveToErrorQueue / dead-lettering is durable-pipeline only, so the registered MoveToErrorQueue step never fires for this chain. What actually happens on exhausted retries is what the controller comment correctly describes: the exception escapes the dispatch and is caught in WorkoutLogsController.EvaluateAdaptationAsync. The two handler comments and the controller comment make incompatible claims about the same failure path; the dead-letter wording would mislead someone debugging a conflict into checking the error queue.

**Suggested fix:** Reword both handler comments to: retries inline a bounded number of times; on exhaustion the conflict propagates to the synchronous caller (the controller maps it to a retryable error envelope) — dead-lettering applies only if this command is ever moved to a durable pipeline. Keep MoveToErrorQueue only if the durable path is genuinely anticipated, and say so.

---
*comment_accuracy | confidence 65% | blind-challenge 90/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
```diff
@@ -0,0 +1,600 @@
...
+    /// append conflict — a rule keyed on it never fires for the adaptation race.
+    /// </remarks>
+    /// <param name="policies">The failure-policy surface to register on.</param>
+    internal static void ConfigureFailureRules(IWithFailurePolicies policies)
+    {
+        ArgumentNullException.ThrowIfNull(policies);
+        policies.OnException<EventStreamUnexpectedMaxEventIdException>()
+            .RetryTimes(MaxConcurrencyRetries)
+            .Then.MoveToErrorQueue();
+    }
```
leehopper, line 368: **💡 [LOW] Bounded-retry-then-marker-replay concurrency semantics are not pinned by any test**

The chain-scoped policy in ConfigureFailureRules is the mechanism that turns a lost append race into a graceful replay (loser retries, hits the winner marker via SeenAsync, returns the replayed envelope). Neither test layer would catch its silent failure. The unit test ConfigureFailureRules_RegistersExactlyOneBoundedRetryRule asserts only that ONE rule exists — not which exception it matches, that it retries MaxConcurrencyRetries times, or that it then dead-letters. The integration race test (ConcurrentEvaluations_SameLogSamePlan_ExactlyOneAdaptationWins) explicitly tolerates the loser EITHER replaying the envelope OR escaping with a ConcurrencyException: envelopes.Should().NotBeEmpty() passes with just the winner, and an escaped conflict is captured as a legal outcome. So if the rule were re-keyed to the wrong exception type — exactly the bug commit d0bfd4a fixed after the original rule targeted Marten.Exceptions.ConcurrentUpdateException — the retry would never fire, the loser would throw instead of replaying, and the entire suite would still pass.

**Suggested fix:** Strengthen the integration race assertion: since the winner commits its marker before the loser retries, the bounded retry deterministically resolves the loser to the replayed envelope — assert envelopes.Should().HaveCount(2) (both outcomes are Adapted/Restructure) and that no conflict escaped. Alternatively (or additionally) make the unit test inspect the registered rule for the EventStreamUnexpectedMaxEventIdException match and the MaxConcurrencyRetries count.

---
*test_coverage | confidence 63% | blind-challenge 92/100 | origin: surfaced | deep-review*

backend/src/RunCoach.Api/Modules/Training/Adaptation/EscalationClassifier.cs
```diff
@@ -36,15 +42,47 @@ public EscalationDecision Classify(
...
+            // re-fires the restructure once the cooldown has elapsed, even while the
+            // rolling score still sits inside the dead-zone — a genuinely
+            // deteriorating runner must not be silently absorbed.
+            if (!cooldownActive && newStreak >= AdaptationThresholds.ConsecutiveMissedDaysForRestructure)
+            {
+                return new EscalationDecision(
+                    EscalationLevel.Restructure,
+                    AdaptationKind.Restructure,
+                    new AdaptationSignalState(PlanState.NeedsAdjustment, newScore, newStreak, deviation.OccurredOn));
+            }
```
leehopper, line 61: **💡 [LOW] Dead-zone hard-trigger duplicates the restructure decision construction; mirror the existing Absorb helper**

The new hard-trigger carve-out (lines 60-63) constructs new EscalationDecision(EscalationLevel.Restructure, AdaptationKind.Restructure, new AdaptationSignalState(PlanState.NeedsAdjustment, newScore, newStreak, deviation.OccurredOn)) — byte-for-byte the same four-line construction as the original enter-threshold path at lines 103-106. Both express the single concept fire a restructure now and stamp the date; if the stamped state ever changes, the two sites can silently diverge. The class already has a one-line Absorb(state) helper for exactly this shape of repetition, so a sibling Restructure(newScore, newStreak, occurredOn) helper keeps the two call sites to one expression each and makes the carve-out read as the same decision the normal path makes.

**Suggested fix:** Before (twice): return new EscalationDecision(EscalationLevel.Restructure, AdaptationKind.Restructure, new AdaptationSignalState(PlanState.NeedsAdjustment, newScore, newStreak, deviation.OccurredOn)); After: add private static EscalationDecision Restructure(double newScore, int newStreak, DateOnly occurredOn) => new(EscalationLevel.Restructure, AdaptationKind.Restructure, new AdaptationSignalState(PlanState.NeedsAdjustment, newScore, newStreak, occurredOn)); and call return Restructure(newScore, newStreak, deviation.OccurredOn); at both sites.

---
*simplification | confidence 57% | blind-challenge 75/100 | origin: new | deep-review*

leehopper: Fixed in 0ecf25b: extracted `private static EscalationDecision Restructure(double newScore, int newStreak, DateOnly occurredOn)`; both the enter-threshold crossing and the dead-zone consecutive-missed-days hard trigger now call it, so the two sites can't silently diverge.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper: Fixed in fda3941: the escalated unswappable-L1 path now persists a restructure-shaped state — `AdaptationSignalState.Create(NeedsAdjustment, score, streak, deviation.OccurredOn)`, derived from the executed level rather than the classifier's MicroAdjust `NextState` — so the cooldown/dead-zone hysteresis engages instead of immediately re-firing. New unit test `Handle_EscalatedMicroAdjust_StampsRestructureStateAndComposesWithRestructureLevel` (fe4806a) asserts the stored document enters `NeedsAdjustment` with `LastAdaptationOn` stamped.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper: Fixed in fda3941: `TryPlanSwap` is now gated on an actual missed key workout (`CompletionStatus == Skipped && IsKeyWorkout`). A score-driven L1 on a completed-but-subpar log absorbs deterministically (state advances, no event) instead of swapping a finished workout (the false "you missed" nudge) or over-escalating a 2.0 signal into an LLM restructure reserved for the 3.0 crossing. Pinned by `Handle_ScoreDrivenMicroAdjustOnACompletedLog_AbsorbsWithNoSwapNoEscalationNoLlm` (fe4806a).

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper: Fixed in fda3941: the post-decode policy now rejects a proposal whose echoed `output.SafetyTier` diverges from the deterministic gate (`safety.Tier`), so GATE-BEFORE-INCREASE no longer depends on the LLM policing its own tier. Pinned by `Handle_LlmEchoesADifferentTierThanTheGate_RejectsTheProposalWithNothingStaged` (fe4806a) — Amber gate + echoed-Green + positive `NetLoadDelta` is rejected with nothing staged.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogsController.cs
leehopper: Fixed in fda3941: the create's never-fail catch widens to `DocumentAlreadyExistsException` (the marker-only paths' duplicate `Insert` race) → the same retryable 201 envelope. `ConcurrentUpdateException` stays uncaught per your challenge nuance: the signal-state document has no optimistic concurrency (last-write-wins `Store`), so it's unreachable for this command and a catch would be dead code. Controller test `CreateLog_DuplicateMarkerConflictEscapesDispatch_MapsToRetryableErrorEnvelope` (fe4806a) covers the duplicate-marker surface through the create path.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper: Fixed in fda3941 + 702b2ef (taking the second option you offered). The handler no longer pre-sanitizes the detail it hands the assembler: it builds one RAW `LoggedWorkoutDetail`, sanitizes a separate copy for the safety gate, and passes the raw detail to `ComposeForAdaptationAsync` — now documented as the single sanitization owner for the prompt (its tests already prove it sanitizes raw input). That removes the triple-escape/double-wrap (handler sanitize was the unaccounted third pass). Pinned by `Handle_RestructurePath_SanitizesViaTheAssemblerOnly_PassingTheRawDetail` (fe4806a). I kept the contract in the parameter doc + test rather than introducing a structural `SanitizedText` type — heavier than this single seam warrants at MVP-0.

leehopper: Fixed in fda3941: `RestructureAsync` now composes with the executed `EscalationLevel.Restructure` on both entry paths (direct L2 and escalated L1), matching the appended event and the validator's hard restructure requirement — no more contradictory "Escalation level: MicroAdjust" in the prompt. The same test as the state fix (fe4806a) asserts `ComposeForAdaptationAsync` receives `Restructure`.

backend/src/RunCoach.Api/Modules/Coaching/ContextAssembler.cs
leehopper: Fixed in 702b2ef: the `ServiceCollectionExtensions` DI comment and the `ContextAssemblerDiResolutionTests` guard doc + assertion reason now describe the 7-arg public constructor (6 required + the optional `IRecentLogSanitizer` that resolves from the registered singleton), keeping the DI-selection rationale accurate.

backend/src/RunCoach.Api/Modules/Coaching/ContextAssembler.cs
leehopper: Documented as deliberate MVP-0 scope in 702b2ef (DEC-080): the runner-profile block is intentionally omitted from `ComposeForAdaptationAsync` at MVP-0 — the `adaptation.v1.yaml` template (landed PR4) defines no profile token, so adding it is a template + eval-cache re-record change out of PR5's scope. The `IContextAssembler`/`ContextAssembler` docs now state this; revisit during Unit 6 eval calibration against the five TestProfiles.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper: Resolved as a documented deliberate decision in 702b2ef (DEC-080): validator rejects are terminal with zero re-prompts at MVP-0. You're right that DEC-073 governs transport-level SDK retries (429/5xx) only and doesn't supersede the spec's re-prompt line — so I recorded the divergence explicitly rather than leaving it implicit. Rationale: a constrained-decoding reject on the frozen schema is unlikely to be fixed by an identical second call, and the clean retryable "try again" envelope already exists; the handler XML doc (fda3941) now records this. Revisit in Unit 6 if eval scores show under-reaction on recoverable rejects.

backend/src/RunCoach.Api/Modules/Training/Workouts/WorkoutLogsController.cs
leehopper: Confirmed deliberate sequencing — `createWorkoutLog`'s Plan/Conversation tag invalidation is assigned to PR7 of this merge train (pr-strategy.md), not PR5, so no change here. The `AdaptationResponseDto` doc describes the target contract PR7 delivers. Leaving this thread for the PR7 unit to close the loop on the frontend wiring.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper: Fixed in fda3941: the `RecordAndReturn` summary now states it records the resolved `Kind=Adapted` response for the given kind and is the shared tail of EVERY committing path — the deterministic absorb/nudge paths AND the LLM restructure path (which calls it with `AdaptationKind.Restructure`) — not "no-plan-change / deterministic only".

backend/tests/RunCoach.Api.Tests/Infrastructure/StubCoachingLlm.cs
leehopper: Fixed in 702b2ef: dropped the `backend/CLAUDE.md` path reference from `StubCoachingLlm`'s remarks. The in-comment explanation (Wolverine codegen service-locates internal types, falls back to scope-based DI, and breaks same-session idempotency) already stands alone.

backend/src/RunCoach.Api/Modules/Training/Adaptation/WorkoutMetricsProjection.cs
leehopper: Fixed in fe4806a: `ToDisplayMetrics_MalformedStoredMetrics_ThrowsJsonException` feeds non-JSON text, a JSON array, and a truncated object and asserts `JsonException` — pinning the documented fail-loudly contract so a future change that wraps the deserialize in a try/catch returning Empty would break the test.

backend/src/RunCoach.Api/Modules/Coaching/IContextAssembler.cs
leehopper: Fixed in 702b2ef: the `triggeringLog` doc is reworded to a current-scope statement — "At MVP-0 this composes the triggering log plus plan-week context only; the recent-logs section carries a single log" — no forward reference to a later unit.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EscalationClassifier.cs
leehopper: Fixed in 0ecf25b: both the class summary and the body comment now read "...where `EvaluateAdaptationHandler` hands off to the coaching LLM" — code-anchored, no PR-number forward reference.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper: Fixed in fda3941: the `MaxConcurrencyRetries` and `Configure` docs now state that on exhaustion the conflict propagates to the synchronous caller (the controller maps it to a retryable envelope) rather than dead-lettering — inline `InvokeForTenantAsync` never reaches the durable error queue (DEC-073 verified). The `MoveToErrorQueue` terminal is kept with an inline comment noting it's inert until/unless this command moves to a durable pipeline. This reconciles with the controller comment.

backend/src/RunCoach.Api/Modules/Training/Adaptation/EvaluateAdaptationHandler.cs
leehopper: Fixed in fe4806a: the concurrency integration assertion is tightened from `NotBeEmpty()` to BOTH invocations resolving to the winner's restructure envelope with no conflict escaping (`outcomes.Should().OnlyContain(o => o.Conflict == null)` + `envelopes.Should().HaveCount(2)`). A re-key of the retry rule to the wrong exception type — the d0bfd4a regression — would now fail this test: the loser would escape as a conflict instead of replaying the winner's marker.

backend/openapi/swagger.json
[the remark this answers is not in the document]
leehopper: Resolving — confirmed deferred, not drift specific to this PR. `AdaptationKind` is C#-nullable on the DTO; the required/non-null marking comes from `RequireNonNullablePropertiesSchemaFilter` promoting every `$ref` enum property repo-wide (40+ shipping properties), and `swagger.json` is generated output behind the codegen drift gate so a hand-edit would fail CI. The structural fix (nullable-aware `$ref` emission) was evaluated and deferred on PR #163.

leehopper:
**Deep-review security finding (metric-value sanitizer bypass)** — the one finding with no inline thread (it targets `RecentLogSanitizer.cs`, outside this PR's diff).

Tracked as a Pre-Public-Release backlog item in 702b2ef (`docs/features/backlog.md`) rather than fixed in PR5. Rationale: it's a defense-in-depth gap, not a break-out (delimiter-escaping + the spotlight nonce still prevent the attacker from closing the section), the root is outside this PR's diff, and the MVP-0 audience is self + family (per DEC-079's high-risk-safety-subset scope). The captured fix is: validate the metrics bag at the create boundary (numeric canonical keys reject non-numeric JSON; reject keys outside `WorkoutMetricKeys.All`) OR route every rendered metric value through the DEC-059 sanitizer rather than only Weather/Terrain.

PR #172, opened 2026-06-10
feat(slice-3): adaptation eval suite + DEC-074 prompt-hash sentinel (PR6)

leehopper:
## Slice 3 PR6 — Unit 6: eval suite + DEC-074 sentinel

Sixth of the 7 dependency-ordered Slice 3 PRs. Builds the adaptation eval suite + the DEC-074 prompt-hash sentinel — and the new eval caught (and this PR fixes) a latent production bug in the adaptation structured-output schema.

### What landed

**Deterministic calibration eval** (`Eval/Adaptation/`, no LLM, always runs):
- 20 classification scenarios across all 5 `TestProfiles` (absorb 8 / nudge 6 / restructure 6, incl. the spec's easy-day in-cap over-performance), driving the *real* `DeviationEngine`→`EscalationClassifier` chain, with **asymmetric scoring** (under-reaction = hard fail; over-reaction = low-not-penalized, DEC-079). Result: zero hard fails, **no threshold tuning needed**.
- **Deterministic `SafetyGate` short-circuit** (9 scenarios): crisis `988`/`741741`, emergency, injury, multi-turn RED-S, benign control. **Safety pass-rate 1.0 ≥ 0.95 gate.**
- Validator-reject + eval-side mileage-ramp constraint (≤10% week-over-week for novel load, with a recovery-ramp exemption up to the recently-held baseline) — enforced against the cached restructure outputs, not just synthetic plans — trademark guard, per-category pass-rate report.

**DEC-074 prompt-hash sentinel:** `Prompts/.prompt-hashes.sha256` manifest + `check-prompt-hashes.sh` + glob-scoped lefthook hook + `EvalTestBase` static-ctor backstop + `rerecord-eval-cache.sh` manifest regen + `.gitattributes` LF pin + 4 dedicated sentinel tests.

**LLM Level-2 restructure + communication judge** (Replay-mode, cache recorded): the real `adaptation.v1.yaml` system prompt → cached Sonnet structured call → asserts validator-passing + Restructure-kind + Green echo → Haiku judge over the rationale (states what/why/trajectory, ≥2 athlete-specific data points, no banned phrasings). Recorded for `lee` + `priya`.

### 🐞 Production bug the eval caught (and this PR fixes)

Recording the L2 restructure cache against the live API surfaced that **`AdaptationSchema.Frozen` — what production `ClaudeCoachingLlm` sends on every L2 restructure — would 400 on its first live call.** `PlanAdaptationOutput` uses `WorkoutOutput[]` in two slots, so `JsonSchemaExporter` emits a `$ref` into `#/properties/...`, and Anthropic constrained decoding rejects any `$ref` not under `$defs`/`definitions`. The adaptation loop was never exercised live before (the live pass is gated after Slice 3), so the eval is the first thing to hit it.

**Fix:** `AnthropicSchemaSanitizer` now inlines every local `$ref` before the schema ships (no-op for ref-free schemas, so existing eval-cache keys stay byte-stable; plan-gen evals send raw `JsonSchemaHelper` output and are unaffected). Regression tests cover the ref-inlining + the `Frozen` no-`$ref` invariant.

### Verification

- **Full suite green: 1705 / 0** (incl. integration).
- **Eval suite replays 81 / 0 with zero API calls** (CI-stable; the 2 restructure + 2 judge fixtures recorded against a funded key, TTL-sentinel'd).
- DEC-074 backstop proven (a prompt-hash mismatch fails the suite).

### Review follow-ups (addressed on this branch)

All 11 deep-review findings + both CodeRabbit comments are addressed in `8e2269a` / `5956ea2` / `d8fc74f`: the `rerecord-eval-cache.sh` DEC-074 deadlock (manifest regen now precedes the Record run), realpath path resolution in both eval scripts, the mileage-ramp guardrail now policing real restructure output (with the recovery-ramp exemption encoded), the easy-day over-performance scenario, a testable `VerifyManifest` backstop core with drift/missing-manifest tests, sanitizer sibling-merge + JSON-Pointer edge tests, removal of the redundant `DeepClone` and the dead `EventCount`, category derivation in the `Classify` factory, and the two doc corrections (eval-results is gitignored; `PlanConstraintEvaluator` is eval-side).

### Follow-ups found

Deferred to the cycle plan's *Captured During Cycle* table: the three LLM-variance-robustness mechanisms from the Unit 6 spec/Gherkin — Cohen's-kappa judge calibration, the per-category 2×SE baseline gate, and pass-2-of-3 — are deferred (all moot in deterministic Replay; revisit when the judged surface grows or the suite runs live).

## Summary by CodeRabbit

* **Tests**
  * Added extensive deterministic adaptation eval suites (classification, safety, restructure, validator, asymmetric scoring, trademark checks) and many supporting scenario, scoring, runner, and report types.
* **Chores**
  * Added prompt-hash sentinel and manifest, verification/regeneration scripts, pre-commit hook, and line-ending handling for prompt files.
* **Bug Fixes**
  * Improved schema sanitizer to inline local JSON Schema $ref references and added related stability tests.
* **Documentation**
  * Updated cycle plan with eval-suite notes.


backend/tests/scripts/rerecord-eval-cache.sh
```diff
@@ -71,9 +71,15 @@ with open('$entry_file', 'w') as f:
 done < <(find "$CACHE_DIR" -name "entry.json" -print0)
 echo "  Patched $PATCHED entry.json files."
 
+# Step 3b: Regenerate the DEC-074 prompt-hash sentinel manifest so the committed
+# cache and the manifest always move together (the manifest records the prompt
+# contents this cache was recorded against).
+echo "[4b/5] Regenerating prompt-hash manifest (DEC-074)..."
+bash "$SCRIPT_DIR/check-prompt-hashes.sh" --write
```
leehopper: **🟠 [HIGH] rerecord-eval-cache.sh deadlocks after a prompt edit (DEC-074 backstop fires before this regen)**

This manifest regeneration runs **too late**. Control flow with `set -euo pipefail`:

1. Step `[2/5]` `rm -rf`'s the entire `$CACHE_DIR`.
2. Step `[3/5]` runs the suite in `EVAL_CACHE_MODE=Record`. That load triggers the new `EvalTestBase` **static constructor**, which calls `VerifyPromptHashManifest()` **unconditionally — no Record-mode escape**.
3. After a `Prompts/*.yaml` edit (the exact drift whose error message tells you to run this script), the recomputed hashes diverge from the still-committed manifest → the static ctor throws → every `Category=Eval` test fails with `TypeInitializationException` → the binary exits non-zero → `pipefail` aborts the script **before this line runs**.

Net: the prescribed remediation can never complete, and the cache is already deleted (recoverable only via `git checkout` + an undocumented manual `check-prompt-hashes.sh --write`). Validated 90%, blind challenge 95%.

**Suggested fix:** move `bash "$SCRIPT_DIR/check-prompt-hashes.sh" --write` to **before** the `[3/5]` Record run (the manifest must reflect the prompts being recorded against anyway), or gate `VerifyPromptHashManifest` on the effective cache mode so Record runs are exempt.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/AdaptationRestructureEvalTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        output.RestructurePlan.Should().NotBeNull();
+
+        // Communication judge: every rationale criterion must pass.
+        verdict.OverallScore.Should().Be(1.0m, because: "the rationale must meet every communication criterion");
+        verdict.Criteria.Should().AllSatisfy(c =>
+            c.Passed.Should().BeTrue(because: $"criterion '{c.CriterionName}' should pass: {c.Evidence}"));
+
+        // Trademark guard on the LLM-authored copy.
+        output.Rationale.Should().NotContainEquivalentOf("VDOT");
+    }
```
leehopper, line 107: **🟡 [MEDIUM] Mileage-jump constraint is never checked against the real LLM output — and the committed lee fixture violates it**

Flagged independently by **two agents** (bug-detector + test-analyzer). `AdaptationConstraintEvaluator` was built specifically to police the 10% week-over-week volume-jump guardrail that constrained decoding and the production validator deliberately do **not** enforce — but it is only ever called against hand-built synthetic plans in `AdaptationValidatorEvalTests`. This restructure eval — the **only** place real (cached) LLM output flows through the suite — never calls `AdaptationConstraintEvaluator.Evaluate(output.RestructurePlan, ...)`.

The gap is not hypothetical: the committed `lee` fixture proposes `revised_weekly_targets` **32 → 36 → 40 km** (+12.5% and +11.1%; the fixture's own `forward_path` calls it "about a 12% bump"). Both jumps exceed the `1.10` ceiling, yet the suite is green — the guardrail the suite advertises is violated by the very output it certifies. Validated 90%, blind challenge 97%.

**Suggested fix:** call `AdaptationConstraintEvaluator.Evaluate(output.RestructurePlan!, baselineWeeklyKm)` here and assert no violations (re-record so the proposal complies) — **or**, if a recovery ramp back toward recently-held volume is intentionally exempt from the 10% rule, encode that exemption in the evaluator and its docs so the rule matches the coaching policy. As shipped, the rule and the gated output contradict each other silently.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/EvalTestBase.cs
```diff
@@ -377,11 +433,32 @@ protected virtual ValueTask DisposeAsyncCore()
...
+        var committed = File.ReadAllText(manifestPath).Replace("\r\n", "\n").TrimEnd('\n');
+
+        if (!string.Equals(actual, committed, StringComparison.Ordinal))
+        {
+            throw new InvalidOperationException(
+                "DEC-074: a prompt under Prompts/ changed but the committed eval cache was not "
+                + "re-recorded — the cache no longer matches the prompt contents. "
+                + "Fix: run backend/tests/scripts/rerecord-eval-cache.sh.");
+        }
+    }
```
leehopper, line 474: **🟡 [MEDIUM] DEC-074 backstop throw path is untested; the PromptEdit sentinel test is tautological**

`VerifyPromptHashManifest`'s two failure branches (missing manifest, drift mismatch) are private, reachable only via the static constructor, and never exercised. `Dec074PromptHashSentinelTests` covers the happy path and format, but `PromptEdit_ChangesTheRecordedHash_SoTheSentinelWouldFire` only asserts that hashing different bytes yields different hashes — a property of SHA-256 that holds regardless of any sentinel code — and *infers* the backstop would fire. If the comparison regressed (equality inverted, normalization swallowing a real diff, or the static-ctor call removed), every test stays green while the backstop silently stops protecting against a stale eval cache — the exact silent-pass DEC-074 exists to prevent. Validated 72%, blind challenge 88%.

**Suggested fix:** extract a testable core `internal static void VerifyManifest(string computed, string committed)` (normalization + ordinal compare + throw), call it from `VerifyPromptHashManifest`, and add tests asserting it (a) throws on a one-line drift, (b) does not throw on CRLF/trailing-newline-only differences, (c) throws when the manifest is missing.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/AdaptationScenarioLibrary.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// </summary>
+    internal static IReadOnlyList<EscalationScenario> ClassificationScenarios { get; } =
+    [
+        Classify("absorb.sarah.ontarget-easy", AdaptationEvalCategory.Absorb, "sarah", EscalationLevel.Absorb, Step(Easy, OnTarget, 0)),
+        Classify("absorb.lee.ontarget-tempo", AdaptationEvalCategory.Absorb, "lee", EscalationLevel.Absorb, Step(Tempo, OnTarget, 0)),
+        Classify("absorb.maria.within-tolerance-easy", AdaptationEvalCategory.Absorb, "maria", EscalationLevel.Absorb, Step(Easy, WithinTolerance, 0)),
+        Classify("absorb.james.ontarget-easy", AdaptationEvalCategory.Absorb, "james", EscalationLevel.Absorb, Step(Easy, OnTarget, 0)),
+        Classify("absorb.lee.single-minor-slow", AdaptationEvalCategory.Absorb, "lee", EscalationLevel.Absorb, Step(Easy, MinorSlow, 0)),
+        Classify("absorb.priya.overperform-tempo", AdaptationEvalCategory.Absorb, "priya", EscalationLevel.Absorb, Step(Tempo, OverPerform, 0)),
+        Classify("absorb.maria.overperform-interval", AdaptationEvalCategory.Absorb, "maria", EscalationLevel.Absorb, Step(Interval, OverPerform, 0)),
```
leehopper: **💡 [LOW] Spec's easy-day over-performance scenario (2-3% cap) is absent from the catalog**

The Unit 6 Gherkin names a scenario "Easy-day over-performance produces no upgrade" with Given "a logged workout that beat its prescription on an **easy day** within the 2 to 3 percent cap." The catalog realizes `OverPerform` only on **Tempo** (`absorb.priya.overperform-tempo`) and **Interval** (`absorb.maria.overperform-interval`), and `DeviationIntent.OverPerform` is realized as a 5% distance overshoot 20 s/km past the fast bound — not an easy day within a 2-3% cap. The core invariant (over-performance never upgrades) is covered, so this is low-severity drift, but it is **not listed among the cycle-plan's documented deferrals**. Validated 75%, blind challenge 90%.

**Suggested fix:** add one `absorb.sarah.overperform-easy` scenario with a small in-cap overshoot, or note the substitution alongside the other documented deferrals in the cycle-plan row.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/AnthropicSchemaSanitizer.cs
```diff
@@ -116,14 +146,124 @@ public static class AnthropicSchemaSanitizer
...
+
+                // Carry any sibling keywords on the $ref node (e.g. an injected
+                // description) onto the inlined object — siblings win.
+                if (inlined is JsonObject inlinedObject)
+                {
+                    foreach (var sibling in obj.Where(kvp => kvp.Key != "$ref"))
+                    {
+                        inlinedObject[sibling.Key] = ResolveNode(sibling.Value, root, activePointers);
+                    }
+                }
```
leehopper, line 195: **🟡 [MEDIUM] Sibling-keyword merge onto an inlined $ref target has no isolated test (contested)**

**(Contested — validator 60, challenge 72.)** None of the four new sanitizer unit tests place a sibling key beside `$ref`, so three behaviors ship without an isolated fixture: (a) siblings are carried at all, (b) a sibling overrides the same key on the inlined target (the documented "siblings win"), and (c) a sibling value that itself contains a `$ref` is resolved rather than copied verbatim. `PlanAdaptationSchemaStabilityTests` gives *indirect* coverage of (a) via the `[Description]` sibling on the repeated `WorkoutOutput` `$ref`, but (b) is unverified, and a regression that dropped/incorrectly-merged siblings would pass the whole suite. Case (c) is low practical risk today (descriptions are strings).

**Suggested fix:** add a unit test with `{ "$ref": "#/properties/primary", "description": "overridden" }` asserting siblings-win precedence on the inlined object, plus one case where a sibling value contains a nested `$ref` asserting the materialized schema is reference-free.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/AnthropicSchemaSanitizer.cs
```diff
@@ -116,14 +146,124 @@ public static class AnthropicSchemaSanitizer
...
+            switch (current)
+            {
+                case JsonObject obj when obj.TryGetPropertyValue(segment, out var child):
+                    current = child;
+                    break;
+                case JsonArray array when int.TryParse(segment, out var index) && index >= 0 && index < array.Count:
+                    current = array[index];
+                    break;
+                default:
+                    return null;
```
leehopper, line 266: **🟡 [MEDIUM] ResolvePointer edge paths untested: array-index, ~0/~1 unescaping, unresolvable-throw (contested)**

**(Contested — validator 45, challenge 72; mostly unreachable from today's schemas.)** Three `ResolvePointer` behaviors have no coverage: array-index segments (`#/anyOf/0`), the `~1`→`/` / `~0`→`~` unescaping (order-sensitive), and the null-return → `Unresolvable $ref` throw (the string "Unresolvable" appears nowhere in the test project). Today's production schemas emit `#/properties/Name` pointers from plain C# identifiers, so the unescaping and unresolvable paths are defensive code for hypothetical future shapes — though the array-index branch may be live via `anyOf`-wrapped nullable types. Low priority, but the branches are real and untested.

**Suggested fix:** add focused tests for an array-index pointer, a `~0`/`~1`-escaped property name, and an external/dangling pointer asserting the `InvalidOperationException`.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/AdaptationConstraintEvaluator.cs
```diff
@@ -0,0 +1,50 @@
+using RunCoach.Api.Modules.Coaching.Adaptation;
+
+namespace RunCoach.Api.Tests.Modules.Coaching.Eval.Adaptation;
+
+/// <summary>
+/// Eval-side deterministic constraint evaluator for a restructure proposal,
+/// mirroring the production <c>PlanConstraintEvaluator</c> for plan generation.
```
leehopper: **💡 [LOW] Doc comment calls PlanConstraintEvaluator "production" but it is test-project infrastructure**

The XML doc says this class mirrors "the **production** `PlanConstraintEvaluator` for plan generation," but `PlanConstraintEvaluator` lives only in the test project (`backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/PlanConstraintEvaluator.cs`) — `backend/CLAUDE.md` lists it under test Eval infrastructure. A reader sent to `backend/src` will not find it. Validated 85%, blind challenge 95%.

**Suggested fix:** reword to "mirroring the existing eval-side `PlanConstraintEvaluator` used for plan generation."

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/AnthropicSchemaSanitizer.cs
```diff
@@ -116,14 +146,124 @@ public static class AnthropicSchemaSanitizer
...
+            case JsonObject obj when TryGetRefPointer(obj, out var pointer):
+                if (!activePointers.Add(pointer))
+                {
+                    throw new InvalidOperationException(
+                        $"Recursive $ref '{pointer}' cannot be inlined for Anthropic constrained decoding.");
+                }
+
+                var target = ResolvePointer(root, pointer)
+                    ?? throw new InvalidOperationException($"Unresolvable $ref '{pointer}' in schema.");
+                var inlined = ResolveNode(target.DeepClone(), root, activePointers);
```
leehopper: **💡 [LOW] Redundant DeepClone of the $ref target before ResolveNode rebuilds it**

`ResolveNode` is a pure rebuilder — every switch arm returns a freshly constructed/detached node and never mutates its input, and the sibling merge writes only into the freshly built `inlinedObject`. So `target.DeepClone()` clones the entire referenced subtree only to have `ResolveNode` immediately rebuild it node-by-node (a second full copy), and it misleads readers into thinking resolution mutates the target in place. Passing `target` directly yields a byte-identical tree (existing tests still pass). Validated 70%, blind challenge 90%.

```suggestion
                var inlined = ResolveNode(target, root, activePointers);
```

leehopper (commented):
## Deep Review — PR #172 (adaptation eval suite + DEC-074 sentinel)

Multi-agent review (7 discovery agents on **Fable**; **Sonnet** for validation + blind challenge). **11 findings** posted inline below: 1 high, 4 medium, 6 low.

**Headline:** the one production change — `AnthropicSchemaSanitizer.ResolveReferences` (`$ref` inlining) — was scrutinized by every agent and **held up** (recursion guard, sibling merge, pointer resolution, and byte-stability are all sound). The actionable findings are in the test/tooling layer:

- 🟠 **High** — the documented recovery script `rerecord-eval-cache.sh` deadlocks after a prompt edit (the new `EvalTestBase` static-ctor backstop fails the `Record` run before the manifest is regenerated).
- 🟡 **Medium** — the 10% mileage-jump guardrail (`AdaptationConstraintEvaluator`) is never applied to the real LLM restructure output, and the committed `lee` fixture (32→36→40 km) violates it while the suite stays green.

**Pipeline notes (transparency):** the blind challenge **removed one finding** (a diamond-`$ref` coverage claim — refuted because `PlanAdaptationSchemaStabilityTests.Frozen_ContainsNoReference` already exercises that path). Two findings (the script deadlock and the `PlanConstraintEvaluator` doc-label) were **rescued from a false-positive injection filter** that matched legitimately-quoted reviewed code (`rm -rf`, a slash-containing file path). Three findings are flagged **(contested)** — lower-confidence, included for author judgement.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/AdaptationEvalReport.cs
```diff
@@ -0,0 +1,67 @@
+namespace RunCoach.Api.Tests.Modules.Coaching.Eval.Adaptation;
+
+/// <summary>
+/// Accumulates scored adaptation scenarios and exposes the suite-level gates
+/// (Unit 6): per-category pass-rates (absorb/nudge/restructure/safety), the
+/// safety pass-rate that gates the PR (≥ 95%, DEC-079), and whether any scenario
+/// hard-failed (an under-reaction / missed safety signal). The accumulated rows
+/// and aggregates serialize into the committed eval-results proof artifact.
```
leehopper: **💡 [LOW] Doc says the eval-results proof artifact is "committed" but eval-results/ is gitignored**

This comment (and `AdaptationEvalSuiteTests.cs:12-13`) says the report serializes into the "**committed** eval-results proof artifact," but `EvalTestBase.GetOutputDirectory()` targets `backend/tests/eval-results/`, which `.gitignore` excludes ("generated outputs, not committed"). The artifact is working-tree-only — "committed" is factually wrong and could lead a reader to expect a tracked baseline that does not exist (notably, the committed-baseline regression gate was explicitly deferred). Validated 90%, blind challenge 97%.

**Suggested fix:** drop "committed" — e.g. "the eval-results proof artifact (working-tree-only; eval-results/ is gitignored)" — in both files.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/EscalationScenarioRun.cs
```diff
@@ -0,0 +1,17 @@
...
+/// resolved on the final step, the count of non-absorb decisions across all
+/// steps, and the per-step levels for diagnostics.
+/// </summary>
+/// <param name="FinalLevel">The escalation level resolved on the last step.</param>
+/// <param name="EventCount">How many steps resolved above <see cref="EscalationLevel.Absorb"/>.</param>
+/// <param name="StepLevels">The level resolved at each step, in order.</param>
+internal sealed record EscalationScenarioRun(
+    EscalationLevel FinalLevel,
+    int EventCount,
+    IReadOnlyList<EscalationLevel> StepLevels);
```
leehopper, line 15: **💡 [LOW] EventCount is computed and threaded through the run loop but never read**

A backend-wide grep shows `EventCount` has only its record definition + XML doc; nothing reads it. `AdaptationEvalSuiteTests.BuildReport` reads only `FinalLevel`; the rapid-fire test derives its own count via `run.StepLevels.Count(l => l == Restructure)`. Yet `EscalationScenarioRunner.Run` carries a mutable `eventCount` local + a per-step `if`-block solely to populate it. Validated 80%, blind challenge 97%.

**Suggested fix:** remove `EventCount` from the record and delete the counter/`if`-block in `Run` (return `new EscalationScenarioRun(levels[^1], levels)`); any future consumer can derive it from `StepLevels`.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/EscalationScenario.cs
```diff
@@ -0,0 +1,27 @@
...
+/// <param name="SafetyTier">The safety tier in force (Green for pure classification scenarios).</param>
+/// <param name="ExpectedLevel">The ground-truth escalation level for the final step.</param>
+/// <param name="Steps">The logged workouts, in order.</param>
+internal sealed record EscalationScenario(
+    string Id,
+    AdaptationEvalCategory Category,
+    string ProfileName,
+    SafetyTier SafetyTier,
+    EscalationLevel ExpectedLevel,
+    IReadOnlyList<EscalationScenarioStep> Steps);
```
leehopper, line 27: **💡 [LOW] EscalationScenario carries a redundant, unenforced Category field (contested)**

**(Contested — validator 40, challenge 60; latent, not an active bug.)** The record stores both `Category` and `ExpectedLevel`, value-identical for classification scenarios (0/1/2), but neither the record nor `AdaptationScenarioLibrary.Classify` enforces agreement. A future catalog entry tagged with a mismatched `Category` would pass the per-scenario exact-match assertion while `AdaptationEvalReport` silently mis-buckets the row — and a stray `Category=Safety` would inflate the denominator of the DEC-079-gated `SafetyPassRate`. Secondary: `Classify`'s `params steps` lets a zero-step scenario compile, which `Run` then throws on at `levels[^1]`. No current entry triggers either.

**Suggested fix:** derive `Category` from `ExpectedLevel` (`(AdaptationEvalCategory)(int)ExpectedLevel`), or validate agreement + non-empty steps in the `Classify` factory so a mis-authored entry fails loudly.

backend/tests/scripts/check-prompt-hashes.sh
```diff
@@ -0,0 +1,72 @@
...
+#
+# Usage:
+#   check-prompt-hashes.sh            # verify the committed manifest is in sync (exit 1 on drift)
+#   check-prompt-hashes.sh --write    # regenerate the committed manifest (called by rerecord-eval-cache.sh)
+
+export LC_ALL=C
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+BACKEND_DIR="$(cd "$SCRIPT_DIR/../.." && pwd)"
+PROMPTS_DIR="$BACKEND_DIR/src/RunCoach.Api/Prompts"
```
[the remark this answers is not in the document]
leehopper: Fixed in 8e2269a with the proposed realpath resolution — applied to both this script and the outside-diff instance in rerecord-eval-cache.sh (lines 13–14). The later `cd` calls now only ever receive absolute paths.

backend/tests/scripts/rerecord-eval-cache.sh
leehopper: Fixed in 8e2269a — the manifest regen moved to step [3/6], before the Record run, so the EvalTestBase static-ctor check sees a manifest that matches the prompts being recorded against and the script can no longer deadlock after a prompt edit.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/AdaptationRestructureEvalTests.cs
leehopper: Addressed in 5956ea2 via the exemption route: AdaptationConstraintEvaluator now encodes the recovery-ramp exemption — ceiling = max(prev × 1.10, recently-held baseline) — since rebuilding toward volume the athlete recently held is re-acclimatization, not novel load; growth past the baseline stays rate-limited at +10%. The restructure eval now calls Evaluate(output.RestructurePlan, baseline) against the cached output for both profiles (lee 40 km, priya 60 km) and asserts no violations, so real LLM output is finally policed. The lee fixture's 32→36→40 ramp complies under the exemption (never exceeds the held 40 km), no re-record needed. Two new validator-eval tests pin the boundary: ramp back to baseline is clean, ramp past it is flagged.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/EvalTestBase.cs
leehopper: Fixed in d8fc74f — extracted internal static EvalTestBase.VerifyManifest(computed, committed?, manifestPath) carrying the normalization, ordinal compare, and both throw branches; the static ctor now delegates to it. Three new tests cover: a one-line drift throws, CRLF/trailing-newline-only differences do not, and a missing manifest (null committed) throws.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/AdaptationScenarioLibrary.cs
leehopper: Added in 5956ea2 — absorb.sarah.overperform-easy realizes a new OverPerformInCap intent (+3% distance at an in-band pace near the fast bound) on an Easy day with ground truth Absorb, matching the Gherkin's 2–3% cap scenario. Catalog is now 20 scenarios (absorb 8), inside the 18–24 envelope.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/AnthropicSchemaSanitizer.cs
leehopper: Added in d8fc74f — ResolveReferences_SiblingKeywordsWinOverTheInlinedTarget puts a description on both the $ref node and its target and asserts the sibling wins while non-conflicting target keys carry; ResolveReferences_ResolvesARefNestedInsideASiblingValue asserts a sibling value containing its own $ref is resolved (materialized schema is reference-free), not copied verbatim.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/AnthropicSchemaSanitizer.cs
leehopper: Added in d8fc74f — an array-index segment test (#/anyOf/0), a ~1/~0 unescaping theory (properties named a/b and a~b), and an unresolvable theory covering both a dangling local pointer and an external URL pointer, asserting the 'Unresolvable $ref' InvalidOperationException.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/AdaptationConstraintEvaluator.cs
leehopper: Fixed in 5956ea2 — reworded to 'mirroring the existing eval-side PlanConstraintEvaluator used for plan generation'.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/AnthropicSchemaSanitizer.cs
leehopper: Fixed in d8fc74f — target is passed directly with a comment noting ResolveNode is a pure rebuilder. Full suite green (1705/0) with the committed replay cache, confirming byte-identical output.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/AdaptationEvalReport.cs
leehopper: Fixed in 5956ea2 — both AdaptationEvalReport and AdaptationEvalSuiteTests now describe the proof artifact as working-tree only (eval-results/ is gitignored).

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/EscalationScenarioRun.cs
leehopper: Removed in 5956ea2 — EscalationScenarioRun is now (FinalLevel, StepLevels) and the runner's counter and if-block are gone; the record doc points future consumers at StepLevels.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/Adaptation/EscalationScenario.cs
leehopper: Addressed in 5956ea2 — Classify now derives Category from ExpectedLevel via an explicit switch (levels outside the classification ladder throw, so a stray Safety bucket is impossible from the factory) and rejects zero-step scenarios at catalog construction.

PR #174, opened 2026-06-11
feat(slice-3): read-only explain-the-change chat panel (pr7)

leehopper:
## Summary

Slice 3 PR7 of 7 (spec 17 § Unit 7) — the last PR of the adaptation loop. The home surface now renders the coach's read-only "Explain-the-change" panel over the PR3 `GET /api/v1/conversation/turns` endpoint:

- **Conversation domain model** (`conversation.model.ts`) — hand-written wire contract over the PR3 endpoint, the `plan.model.ts` convention: int-enum const maps (`CONVERSATION_ROLE`, `ESCALATION_LEVEL`, `SAFETY_TIER`, `REFERRAL_CATEGORY`, `ADAPTATION_KIND`) and a turn union discriminated on `role` with the accurate nullability the generated zod cannot express (`escalationLevel`/`adaptationKind`/`diff` null on safety turns — the known `RequireNonNullablePropertiesSchemaFilter` gap). The diff's `before`/`after` reuse `MicroWorkoutCardDto` (same `WorkoutOutput` wire shape).
- **RTK wiring** — new `Conversation` tag + `getConversationTurns` query; `createWorkoutLog` now invalidates `WorkoutLog` + `Plan` + `Conversation`, so a log that synchronously adapts the plan refetches the plan view and the panel in the same interaction (no create-response coupling).
- **Panel components** (`modules/coaching/`) — per OI-5: absorb is silent (no turn, no empty shell), nudge renders as a quiet inline one-liner, restructure as an expandable block with a collapsible **"Show what changed"** structured before/after diff (workout swaps + weekly volume edits, never parsed prose). Safety turns always render in full; the Red crisis turn carries the strongest accent and complete scripted content. No input affordance of any kind (Slice 4). Semantic tokens + `motion-reduce:` pairing throughout.
- **`--warning` semantic token** — new amber severity accent (Catppuccin peach) used only for the subtle left-edge border. Like `--border` it is a supplementary non-text indicator (WCAG 1.4.11-exempt — severity is always conveyed by content/structure), so it is deliberately not added to the check-contrast PAIRS matrix; documented in `index.css`. Red accents reuse the gated `--destructive`.
- **Panel placement** — between `TodayCard` and `UpcomingList`: the explanation refers to the latest log, so it reads in today's context before the forward-looking sections.

### Scope note: Unit 7's mapper item already shipped

The spec's "net-new `WorkoutLog → LoggedWorkoutDetail` mapper" item shipped in PR5 (#169): `EvaluateAdaptationHandler.BuildDetail` + `WorkoutMetricsProjection.ToDisplayMetrics` + DEC-059 `RecentLogSanitizer` wiring on both the safety-gate and prompt paths, with test coverage. Verified against source before building — **this PR is frontend + E2E only; zero backend changes** (so no OpenAPI/codegen drift).

### Fix: antiforgery seed race (pre-existing, root-caused in-PR)

Writing the PR7 E2E surfaced an intermittent full-suite failure that reproduces on clean `main` (~1 in 3 full runs): the register POST intermittently 400s and the SPA stays on `/register`. The Playwright trace shows the mechanism — the app-boot `GET /xsrf` is fire-and-forget (`useAuthBootstrap`), and a fast first submit outran it (POST fired ~6ms before the seed response; no `X-XSRF-TOKEN` header on the request). This is an app-level race (a fast user on a slow connection hits the same 400), not a test artifact, so per the no-flaky-tests rule it got a real fix rather than a retry: `baseQueryWith401Handler` now lazily seeds the pair when a mutation fires with no SPA-readable cookie (idempotent 204; failed seed falls through to the mutation's own error). Covered by four new base-query specs; mutation-dispatching api specs seed the cookie in their harness to mirror the booted runtime. **8/8 consecutive full E2E suite runs green after the fix** (before: ~1 in 2 failed on this branch, 2 of 6 on main).

## Test plan

- [x] `npm run test` — 485 passed (panel/diff/api/page specs per the Unit 7 proof artifacts: silent absorb, inline nudge, expandable restructure + diff collapse/expand, amber/red accents, Red full prominence incl. exact 988/741741 strings, no input affordance, banned-phrasing + trademark guard, motion-reduce pairing, Plan+Conversation invalidation refetch)
- [x] `npm run lint` / `npm run build` / `npx tsx scripts/check-contrast.ts` (28/28 pairs) — green
- [x] Playwright E2E on the host-run stack — `adaptation-panel.spec.ts` (nudge journey + wire-stubbed restructure journey with diff expand/collapse, request-order proof of same-interaction refetch); full suite 15/15 × 8 consecutive runs

### Follow-ups found

- [ ] The panel renders the active plan's full turn history unpaginated — fine for MVP-0 self+family volumes; revisit sizing/pagination when Slice 4 reworks the panel into the interactive chat surface.

## Summary by CodeRabbit

* **New Features**
  * Conversation panel added to Home showing coach adaptations (nudges, restructures) and safety alerts.
  * Collapsible "what changed" before/after diff displays workout and weekly target changes.

* **Behavior**
  * Home plan re-renders to show adjusted plan after submitting a deviating workout; conversation content updates accordingly.

* **Tests**
  * End-to-end and component tests cover panel, diff expand/collapse, nudge vs restructure behavior, and safety-tier rendering.


frontend/src/app/modules/coaching/models/conversation.model.ts
```diff
@@ -0,0 +1,187 @@
...
+// member required because the backend schema filter cannot express C#
+// nullability on `$ref` properties; THIS file is the accurate contract:
+// `escalationLevel`, `adaptationKind`, and `diff` are null on safety turns
+// and always present on adaptation turns, which makes `role` a discriminant
+// (mirrors `ConversationTurnView.FromAdaptation` / `FromSafety`).
+//
+// Renaming or reordering members on either side of the wire requires a paired
+// change in this file plus the backend record. Per the trademark rule in the
+// root `CLAUDE.md`, every user-facing string this module renders must use
+// "Daniels-Gilbert zones" / "pace-zone index" terminology.
```
leehopper, line 23: **🟡 [MEDIUM] Hand-written conversation wire contract has no automated drift tripwire — all tests are self-referential**

The integer-enum contract here (`CONVERSATION_ROLE`, `SAFETY_TIER`, `ADAPTATION_KIND`, …) matches the backend today (verified member-for-member), but every test layer re-encodes the same hand-assumed integers: the unit fixtures are built from these const maps, and the E2E spec deliberately re-declares them inline. The committed generated schema at `api/generated/zod/conversation/conversation.ts` — kept in sync with the backend by the codegen drift gate — is imported by zero specs. If the backend contract drifts (member renumbered, property added/removed, nullability changed), codegen regenerates the zod file and every frontend test stays green while the panel misroutes turns: a `role` flip would send a safety turn (`diff: null`) into the restructure branch, crash `BeforeAfterDiff`, and `AppErrorBoundary` would blank the home page — defeating both the crisis-content surface and the tested “supplementary surface never blocks the plan view” guarantee.

**Suggested fix:** Add a contract spec (e.g. `conversation.model.spec.ts`) that wraps the shared fixtures (`buildNudgeTurn`, `buildRestructureTurn`, `buildCrisisSafetyTurn`, `buildAmberSafetyTurn`) in a `ConversationTurnsResponseDto` and asserts the generated zod response schema's `safeParse` succeeds — turning the codegen drift gate into a tripwire that forces the paired model update this file's header promises. The blind challenger (score 90) also noted the bidirectional exhaustiveness-guard pattern already used for `SuggestedInputType`/`CompletionStatus` in `generated/index.ts` was not applied to the conversation enums.

*Confidence: 60% | dimension: test-coverage | deep-review*

frontend/src/app/api/conversation.api.ts
```diff
@@ -0,0 +1,24 @@
...
+//
+// `getConversationTurns` is tagged `Conversation` so the `createWorkoutLog`
+// mutation can declare `invalidatesTags: [..., 'Conversation']` and the
+// read-only "Explain-the-change" panel refetches in the same interaction as
+// the plan view (spec 17 § Unit 7).
+export const conversationApi = apiSlice.injectEndpoints({
+  endpoints: (builder) => ({
+    getConversationTurns: builder.query<ConversationTurnsResponseDto, undefined>({
+      query: () => ({ url: '/v1/conversation/turns', method: 'GET' }),
+      providesTags: ['Conversation'],
```
leehopper, line 18: **💡 [LOW] New Conversation cache is never cleared on auth-lifecycle changes**

`getConversationTurns` provides the new `Conversation` tag, but the only invalidator anywhere is `createWorkoutLog`. The auth endpoints invalidate only `Auth`, the `loggedOut` reducer only flips slice state, and nothing in the repo calls `apiSlice.util.resetApiState()`. On session expiry (401 → `loggedOut` → `/login`), cached turns survive the `keepUnusedDataFor` window (default 60s); a different account logging in within that window remounts `HomePage`, resubscribes the same cache key, and is served the prior account's turns — including crisis/safety content — without a refetch.

This lifecycle gap pre-dates the PR (the `Plan`/`WorkoutLog` caches share it); this PR adds a more sensitive dataset to it. Contested in review: the 60-second shared-device window is narrow and multi-account use is outside the MVP-0 self+family deployment model — so this reads as a pre-public-release structural item, not a blocker.

**Suggested fix:** Dispatch `apiSlice.util.resetApiState()` on auth transitions (e.g. an `extraReducers` listener on `loggedOut` and login/register fulfilled) so all per-user caches — `Plan`, `WorkoutLog`, `Onboarding`, `Conversation` — are purged on user switch.

*Confidence: 30% (contested) | dimension: cross-file impact | deep-review*

frontend/src/app/api/conversation.api.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+class PatchedRequest extends OriginalRequest {
+  constructor(input: RequestInfo | URL, init?: RequestInit) {
+    if (typeof input === 'string' && input.startsWith('/')) {
+      super(new URL(input, 'https://localhost:5173').toString(), init)
+      return
+    }
+    super(input, init)
+  }
+}
+
```
leehopper: **💡 [LOW] PR adds 3rd and 4th identical copies of the PatchedRequest jsdom workaround plus more hardcoded xsrf cookie-seed harness**

The byte-identical `PatchedRequest` class now exists in four spec files (`plan.api.spec.ts`, `workout-log.api.spec.ts` pre-existing; this file and `base-query.spec.ts` new), and the antiforgery cookie seed/clear pair hardcoding the `__Host-Xsrf-Request` literal is repeated across `plan.api.spec.ts` and `workout-log.api.spec.ts` while `base-query.ts` keeps `XSRF_COOKIE_NAME` module-private. If the cookie name ever changes, the literal-based seeds become silently inert, the lazy XSRF seed fires, and call-count assertions fail confusingly; a jsdom workaround tweak must be applied in four places. (Credit where due: the new `base-query.spec.ts` block already uses a named constant + `setCookie`/`clearCookie` helpers — extracting exactly those would solve this.)

**Suggested fix:** Add a shared spec helper module (e.g. `frontend/src/app/api/test-helpers.ts`) exporting `PatchedRequest`, `seedXsrfCookie()`/`clearXsrfCookie()`, and the cookie-name constant (or export `XSRF_COOKIE_NAME` from `base-query.ts`), and replace the four inline classes and the inline cookie seed/clear pairs with imports.

*Confidence: 57% (contested) | dimension: simplification | deep-review*

frontend/src/app/api/base-query.ts
```diff
@@ -63,6 +64,17 @@ export const baseQueryWith401Handler: BaseQueryFn<
...
+  // Lazy antiforgery seed: the app-boot `GET /xsrf` (`useAuthBootstrap`) is
+  // fire-and-forget, so a fast first submit can reach the server before the
+  // boot seed lands — the mutation then carries no double-submit pair and
+  // the antiforgery filter rejects it with 400. When the SPA-readable
+  // cookie is still absent, seed it inline before the mutation goes out.
+  // The endpoint is an idempotent 204, so racing the boot seed is harmless;
+  // a failed seed falls through and the mutation surfaces the real error.
+  if (api.type === 'mutation' && readXsrfCookie() === null) {
+    await rawBaseQuery({ url: '/v1/auth/xsrf', method: 'GET' }, api, extraOptions)
+  }
```
leehopper, line 79: **💡 [LOW] App-wide antiforgery auth fix bundled into the chat-panel feature PR (process note)**

Commit 8021081 changes `baseQueryWith401Handler` — the code path of every RTK Query request in the app — inside a +1675-line feature PR. REVIEW.md's single-concern rule and the root CLAUDE.md mandate for focused manual review of auth code both apply to this kind of change. Mitigations verified during review: the PR body calls the fix out prominently in its own section with the root-cause trace, and it's covered by four new `base-query.spec.ts` cases — so this is a process reminder (a standalone fix PR would have given the auth path its own focused review), not a code defect. The seed logic itself was traced clean: anonymous idempotent 204 endpoint, no recursion, failed seed falls through, no CSRF weakening.

*Confidence: 15% (validator heavily discounted given the PR-body call-out) | dimension: convention | deep-review*

leehopper (commented):
### Deep Review

No blocking issues — 0 critical, 0 high, 0 main-report findings. 4 improvement suggestions (1 medium, 3 low) posted as inline comments:

1. **[test-coverage]** 🟡 Hand-written conversation wire contract has no automated drift tripwire — `conversation.model.ts`
2. **[cross-file]** 💡 New `Conversation` cache never cleared on auth-lifecycle transitions — `conversation.api.ts`
3. **[simplification]** 💡 3rd + 4th copies of the `PatchedRequest` jsdom workaround + hardcoded xsrf cookie-seed harness — `conversation.api.spec.ts`
4. **[convention]** 💡 App-wide antiforgery fix bundled in the feature PR (process note; the PR body does call it out prominently) — `base-query.ts`

**Clean dimensions:** bug-detector and security-reviewer returned zero findings after cross-verifying the hand-written int-enum wire contract field-by-field against the backend records (all five enums match member-for-member; the nullability split and newest-first ordering confirmed) and tracing the lazy antiforgery seed across every mutation path (no recursion, anonymous idempotent 204 seed endpoint, no CSRF weakening, no XSS sink on LLM-authored content).

**Pipeline:** 7 review agents (bugs, security, cross-file impact, tests, conventions/intent, type design, simplification) → deterministic verification → independent validation → blind challenge. 2 findings were eliminated by validation thresholds: an intent note that the adaptation E2E stubs the units under test (validator 60 — the strategy comment partially acknowledges the trade-off) and a type-narrowing suggestion for `AdaptationTurnDto.referralCategory` → `typeof REFERRAL_CATEGORY.none` (validator 50 — advisory).

---
Generated by deep-review | Reviewed up to: 8021081331fc427e06250939207591b06e549699

frontend/src/app/api/workout-log.api.ts
```diff
@@ -33,7 +39,7 @@ export const workoutLogApi = apiSlice.injectEndpoints({
         method: 'POST',
         body,
       }),
-      invalidatesTags: ['WorkoutLog'],
+      invalidatesTags: ['WorkoutLog', 'Plan', 'Conversation'],
```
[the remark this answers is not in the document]
leehopper: Fixed in 0cd39d9. Verified against RTK Query's `invalidationByTags` middleware: it fires on `isAnyOf(isFulfilled, isRejectedWithValue)`, and `fetchBaseQuery` surfaces HTTP errors as `rejectedWithValue` — so the static array did invalidate `Plan`/`Conversation` on a failed create, and the "fires only on a fulfilled mutation" comment was wrong.

Switched to the callback form `(_result, error) => error === undefined ? ['WorkoutLog', 'Plan', 'Conversation'] : []`, corrected the comment, and added a `does not refetch Plan or Conversation when the create fails` test (the create rejects with a 400, i.e. a rejected-with-value base-query error, and both subscription counts stay at 1).

frontend/src/app/modules/coaching/models/conversation.model.ts
leehopper: Addressed in 0cd39d9 with a contract spec (`conversation.model.spec.ts`) that validates the shared fixtures against the generated `GetApiV1ConversationTurnsResponse` Zod schema:

- a fully-populated adaptation turn (`buildRestructureTurn` / `buildNudgeTurn`) must `safeParse` — `strictObject` rejects an added/removed property and the literal unions reject a renumbered enum member, so a backend drift that regenerated the schema fails this spec until the model + fixtures are updated in lockstep;
- a safety turn (`buildCrisisSafetyTurn` / `buildAmberSafetyTurn`) must be rejected — locking in the documented non-nullable `$ref` gap, so a future nullable-aware codegen flips the assertion and prompts a model review.

On the challenger's exhaustiveness-guard note: I didn't extend the `SuggestedInputType`/`CompletionStatus` compile-time guard to these enums because that pattern keys off standalone generated enum schemas, whereas orval emits the conversation enums only as inline unions inside the one nested response schema. The runtime `safeParse` tripwire is the practical equivalent and also covers property add/remove.

frontend/src/app/api/conversation.api.ts
leehopper: Deferring as a pre-public-release structural item — not fixing in this PR. The gap is shared by the `Plan`/`WorkoutLog`/`Onboarding` caches and predates the panel; the fix (`apiSlice.util.resetApiState()` on auth transitions) is an app-wide auth-lifecycle change that's out of scope for the chat-panel feature and only bites under multi-account shared-device use within the 60s `keepUnusedDataFor` window — outside the MVP-0 self+family deployment model. Tracking it for the pre-public-release hardening pass alongside rate-limiting and request size caps.

frontend/src/app/api/conversation.api.spec.ts
leehopper: Addressed in 0cd39d9. Extracted `PatchedRequest` + `seedXsrfCookie`/`clearXsrfCookie` into `frontend/src/app/api/test-helpers.ts` and exported `XSRF_COOKIE_NAME` from `base-query.ts`. Replaced all four inline `PatchedRequest` copies and the hardcoded `__Host-Xsrf-Request` seed/clear literals across `plan.api.spec.ts`, `workout-log.api.spec.ts`, `conversation.api.spec.ts`, and `base-query.spec.ts`. The seeds now flow from the single exported constant, so a cookie rename can't leave them silently inert.

frontend/src/app/api/base-query.ts
leehopper: Acknowledged — no code change. The antiforgery seed fix already has its own PR-body section with the root-cause trace and four `base-query.spec.ts` cases; splitting a committed fix into a standalone PR at this point isn't worth the churn. Noting the single-concern preference for the next auth-path change.

PR #182, opened 2026-06-11
test(slice-3): make stream-collision concurrency test deterministic

leehopper:
## Summary

Root-causes and fixes the `OnboardingTurnConcurrencyIntegrationTests` CI flake (`successCount == 2`, observed once on the coverage-instrumented `Backend analysis` job during PR4 #167, scheduled for dedicated investigation in cycle plan § Captured During Cycle before the Slice 3 live end-to-end validation pass).

**Verdict: hypothesis (b) — test-harness timing artifact. No gap in Marten's Rich-mode stream-collision guarantee. No data corruption, observed or possible.**

## Root cause

The test fired 5 end-to-end handler+commit tasks via `Task.WhenAll` with no synchronization. Its exactly-one-winner assertion silently assumes every handler reads "no stream yet" **before** any task commits. `OnboardingTurnHandler` only calls `StartStream` when `LoadAsync<OnboardingView>` returns null; when the view already exists it takes the **legitimate second-turn Append path and succeeds** (correct production behavior — that's just the next onboarding turn).

On the slow 4-core coverage-instrumented CI runner, the 5 tasks serialized enough that one task started after the winner's commit, read the committed view, appended turn 2, and succeeded → `successCount == 2` with the stream guarantee fully intact.

## Investigation evidence

- **CI forensics:** attempt 1 of run 27218170242 shows the only assertion that fired was `Expected successCount to be 1 … but found 2`.
- **Stress loop (old shape):** 30/30 green locally under coverlet instrumentation — consistent with a scheduling artifact that needs a slow runner, not a reproducible guarantee gap.
- **Straggler probe (deterministic):** invoking the handler once, committing, then invoking again reproduces the "second success" — and the stream dump shows ONE stream row, ONE `onboarding_started_v1` (v1), winner turn events v2–5, straggler turn events v6–9. No duplicate stream, no duplicate version.
- **True-overlap probe:** staging all 5 `StartStream`s before any commit, then racing all 5 `SaveChangesAsync` calls: exactly 1 winner + 4 stream-collision failures in 20/20 rounds.
- **Schema dump (provisioned by Marten 9, conjoined tenancy):** the guard is layered — `mt_streams` PK `(tenant_id, id)`, `mt_events` unique `(tenant_id, stream_id, version)`, plus the events→streams FK. Two committed `StartStream`s for the same user are structurally impossible.

## Fix

Restructure the test into two phases (the cycle plan's sanctioned option (b), strengthened):

1. **Prepare:** run `OnboardingTurnHandler.Handle` for all 5 submissions sequentially, each in its own DI scope + session, **without committing**. Since nothing has committed, every invocation deterministically reads `view == null` and stages `StartStream` — regardless of machine speed or scheduling.
2. **Commit:** race all 5 `SaveChangesAsync` calls. Exactly one stream-row insert can satisfy the `mt_streams (tenant_id, id)` PK; the rest must surface `ExistingStreamIdCollisionException` or a `MartenCommandException` wrapping SqlState 23505.

This is **not** a retry/rerun/loosened assertion: assertions are unchanged (1 winner, 4 collision-typed failures). The change removes the hidden timing assumption and makes the collision path exercised on **every** run — under the old shape it was only exercised when the tasks happened to overlap.

## Verification

- 15/15 green looping the rewritten test under coverlet instrumentation (the flaking CI condition).
- Full backend suite green.
- Temporary diagnostic probes used for the investigation were deleted; nothing committed beyond the test rewrite + docs.

## Docs

- Cycle plan § Captured During Cycle: 2026-06-09 flake row resolved with findings.
- Cycle plan Status + ROADMAP: live end-to-end validation pass no longer gated on this investigation.

## Summary by CodeRabbit

* **Tests**
  * Improved onboarding concurrency integration test to use a deterministic two-phase execution; validates concurrent commit behavior and ensures exactly one successful commit while detecting expected collision failures.
* **Documentation**
  * Updated roadmap and cycle-plan to mark the flaky concurrency investigation as resolved and note that the live end-to-end validation pass is now unblocked.


leehopper:
@coderabbitai review

PR #187, opened 2026-06-12
fix(slice-3b): trademark scrub on all persisted llm prose (f2)

leehopper:
## Summary

Slice 3B **F2** (`docs/plans/mvp-0-cycle/slice-3b-live-pass-fixes.md` § F2): the 2026-06-11 live pass persisted the trademarked pace-index term in `Macro.Rationale` — outside the assembled-prompt guard and the adaptation-rationale guard. This lands the three required defense layers:

1. **Prompt (a)** — `coaching-system.v1.yaml` gains a `VOCABULARY RULES` block naming "Daniels-Gilbert zones" / "pace-zone index" as the only permitted index vocabulary. Worded without the term itself, so the `ContextAssemblerTests` assembled-prompt Theory stays green. (`adaptation.v1.yaml` already carried the equivalent rule.)
2. **Deterministic boundary guard (b)** — new `TrademarkScrubber` (generated regex, case-insensitive, dot/hyphen/plural variants) hooked into `ClaudeCoachingLlm` on the raw response text before deserialization. This is the single chokepoint every `ICoachingLlm` call crosses, so macro/meso/micro, adaptation, onboarding, and any *future* structured-output type are covered structurally rather than per-type. Occurrences are rewritten to "pace-zone index" and warn-logged (source-generated `LoggerMessage`) so prompt drift stays visible in telemetry. A prose-only leak deliberately does **not** terminally fail the call — DEC-080's terminal posture stays reserved for structural defects (F3/F4); a DEC entry recording this posture lands with the slice-status docs PR after merge.
3. **Eval guard (c)** — new `TrademarkProseGuard` (single-sourced on the production regex) scans **every** prose field of the cached plan-generation outputs for all five profiles (+ the james narrative) and the **whole** adaptation output (previously only `Rationale` was scanned).

The prompt edit invalidates the eval cache, so the change rides with a full DEC-074 re-record (`rerecord-eval-cache.sh`): manifest regenerated first, all fixtures re-recorded against a funded key, TTL-extended, replay-verified. `backend/CLAUDE.md`'s trademark paragraph is updated to describe the new scrub layer and the carve-out class for guard tests.

## Verification

- TDD: scrubber tests watched red (11 failing assertions on a stub) then green; adapter scrub tests watched red (2) then green.
- Full backend suite on this branch: **1736/1736 pass** (evals in Replay against the fresh cache; DEC-074 sentinel green).
- `rg -i` for the term: zero hits in `backend/src`; zero hits across the freshly recorded `eval-cache` (the new prompt rule keeps Sonnet clean at the source; the scrub is the runtime backstop).
- The recorded live-leak shape ("…your VDOT sits around 38") is pinned as a fixture in `TrademarkScrubberTests` and `ClaudeCoachingLlmTests`.
- F2 acceptance at the live surface (fresh persisted events) is proven by the slice's gated live-pass re-run, per the slice done-gate.

## Follow-ups found

- DEC-082 (scrub-not-reject posture) + slice/cycle/ROADMAP status updates land in the standard post-merge docs PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Coaching content now removes prohibited trademarked pace-index terms and substitutes approved wording (e.g., "pace-zone index" / "Daniels-Gilbert zones") in both narrative and structured outputs.

* **Tests**
  * Added tests and guards verifying trademark removal across generation, structured deserialization, persisted eval outputs, and JSON string fields.

* **Documentation**
  * Updated coaching prompt and backend guidance to clarify allowed terminology and carve-outs for tests and docs.


leehopper:
@coderabbitai full review

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/TrademarkProseGuard.cs
```diff
@@ -0,0 +1,38 @@
...
+/// — tighten the prompt and re-record; the production scrub remains the
+/// runtime backstop either way.
+/// </summary>
+internal static class TrademarkProseGuard
+{
+    private static readonly JsonSerializerOptions SerializeOptions = new()
+    {
+        PropertyNamingPolicy = JsonNamingPolicy.SnakeCaseLower,
+        Converters = { new JsonStringEnumConverter() },
+    };
```
leehopper: **💡 [LOW] TrademarkProseGuard duplicates StructuredOutputSerializerOptions instead of reusing it**

**💡 [low] [simplification]** TrademarkProseGuard hand-copies the production serializer options instead of reusing `ClaudeCoachingLlm.StructuredOutputSerializerOptions`

The `private static readonly JsonSerializerOptions SerializeOptions` declared here (lines 20–24) is byte-for-byte identical to `ClaudeCoachingLlm.StructuredOutputSerializerOptions` (the field the production path uses before deserialization). The guard’s stated goal is to single-source detection from `TrademarkScrubber` so it can’t drift from production; this copy creates an analogous drift risk on the serialization side — a future options change (e.g. a new converter) silently diverges the scan shape from the actual wire format.

`ClaudeCoachingLlmTests.cs` (added in this same PR) already accesses `ClaudeCoachingLlm.StructuredOutputSerializerOptions` directly, confirming it is reachable from the test assembly via `InternalsVisibleTo`.

**Suggested fix:** Delete the private `SerializeOptions` field (and the `System.Text.Json.Serialization` using directive if it becomes unused), then change the `JsonSerializer.Serialize` call to `JsonSerializer.Serialize(output, ClaudeCoachingLlm.StructuredOutputSerializerOptions)`.

---
*Confidence: 75% | deep-review*

backend/tests/RunCoach.Api.Tests/Modules/Coaching/TrademarkScrubberTests.cs
```diff
@@ -0,0 +1,121 @@
...
+/// <c>ContextAssemblerTests</c> and <c>OnboardingPromptTests</c>, it must spell
+/// the term out to prove it gets scrubbed.
+/// </summary>
+public class TrademarkScrubberTests
+{
+    [Fact]
+    public void Scrub_ReplacesTheTerm_InTheRecordedLiveLeakShape()
+    {
+        // Arrange — the exact prose shape the 2026-06-11 live pass persisted
+        // into Macro.Rationale (slice-3b-live-pass-fixes.md § F2).
```
leehopper: **💡 [LOW] Comment embeds a literal planning-doc filename instead of the symbolic reference**

**💡 [low] [convention]** Code comment references an external doc file (`slice-3b-live-pass-fixes.md`) instead of a symbolic reference

Lines 19–20 contain the literal path `slice-3b-live-pass-fixes.md § F2`. The REVIEW.md rule requires comments to be self-contained; only symbolic Spec / DEC / R-NNN references are allowed. The XML doc on this same file already uses the compliant form `(Slice 3B F2)`. The planning doc may be renamed or archived after the cycle, leaving a dangling reference.

**Suggested fix:** Replace with the symbolic form used everywhere else in the file (see committable suggestion below).

---
*Confidence: 73% | deep-review*

```suggestion
        // Arrange — the exact prose shape the 2026-06-11 live pass persisted
        // into Macro.Rationale (Slice 3B F2).
```

backend/src/RunCoach.Api/Modules/Coaching/TrademarkScrubber.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// original instance when the text is clean, so the hot no-leak path
+    /// allocates nothing.
+    /// </summary>
+    internal static string Scrub(string text, out int occurrences)
+    {
+        occurrences = TrademarkedTerm().Count(text);
+        return occurrences == 0 ? text : TrademarkedTerm().Replace(text, Replacement);
+    }
+
+    [GeneratedRegex(@"\bV[.-]?DOTs?\b", RegexOptions.IgnoreCase | RegexOptions.CultureInvariant)]
```
leehopper, line 62: **💡 [LOW] Scrub regex misses the trademarked term when it follows a JSON escape sequence in the raw structured-output path**

**💡 [low] [bug]** Scrub regex misses the trademarked term after a JSON `\n` escape in the raw structured-output path

`GenerateStructuredAsync` applies `TrademarkScrubber.Scrub` to the **raw JSON document** before deserialization (`ClaudeCoachingLlm.cs` line 210). In raw JSON, a newline inside a string value is the two-character sequence `\n` (backslash + `n`). The regex anchor `\b` requires a non-word-character boundary; because `n` is a word character, there is no boundary between the `n` of the escape and the `V` of the term. Any occurrence immediately preceded by `\n`, `\t`, `\r`, or `\uXXXX` is **not matched, not replaced, and not warn-logged**.

Multi-paragraph prose fields (rationales, week summaries) routinely encode newlines as `\n`, so a rationale like `"Key metrics:\nVDOT: 38"` is a realistic, triggerable input: the term survives the scrub, deserializes into the prose field, and is persisted — exactly the failure this guard exists to prevent. The `GenerateAsync` text path is **unaffected** (operates on decoded text).

**Suggested fix:** Scrub decoded string values rather than the raw escaped document: parse to `JsonNode`, walk every string value, apply `TrademarkScrubber.Scrub` to each decoded value, then deserialize `T` from the scrubbed node. Alternatively, widen the leading anchor to also accept a preceding JSON escape character, and add a regression test whose prose field contains a newline immediately before the term.

---
*Confidence: 72% | deep-review*

leehopper (commented):
## Deep Review: PR #187 — trademark scrub on all persisted llm prose (f2)

Found **5 improvement suggestions** (0 blocking, 0 action items).

All are low-severity. No issues block merge.

| # | Finding | File | Confidence |
|---|---------|------|-----------|
| simplify-1 | TrademarkProseGuard duplicates StructuredOutputSerializerOptions | TrademarkProseGuard.cs:20 | 75% |
| conv-1 | Comment embeds external doc path instead of symbolic ref | TrademarkScrubberTests.cs:20 | 73% |
| bug-1 | Regex misses term after JSON \n escape in raw-JSON scrub path | TrademarkScrubber.cs:47 | 72% |
| bug-2 | AssertClean false-negative for newline-prefixed term in eval guard | TrademarkProseGuard.cs:31 | 72% |
| test-1 | Safety-boundary eval fixtures re-recorded with no trademark guard | coaching-system.v1.yaml:58 | 67% |

---
Generated by deep-review | Reviewed up to: 7cd72bb020d59b89313855d962d9b9ef5d9573e6

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/TrademarkProseGuard.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// is free of the trademarked pace-index term (case-insensitive, separator
+    /// and plural variants included).
+    /// </summary>
+    internal static void AssertClean(string label, object output)
+    {
+        var json = JsonSerializer.Serialize(output, SerializeOptions);
+        TrademarkScrubber.ContainsTrademarkedTerm(json).Should().BeFalse(
+            because: $"every persisted prose field of '{label}' must use 'Daniels-Gilbert zones' / "
+                + "'pace-zone index' and never the trademarked pace-index term (Slice 3B F2)");
+    }
```
leehopper, line 58: **💡 [LOW] AssertClean passes falsely when the trademarked term follows a newline in any prose field**

**💡 [low] [bug]** `TrademarkProseGuard.AssertClean` passes falsely when the term follows a newline in any prose field

`AssertClean` serializes the decoded output object with `JsonSerializer` and runs `ContainsTrademarkedTerm` on the result. `System.Text.Json` must escape control characters, so a real newline in a prose field becomes the two-character sequence `\n` in the serialized text — the same root cause as bug-1, recreated on the eval side by the serialize step. A cached fixture whose rationale or narrative contains the trademarked term at the start of a line **passes the guard with a false negative**, so the eval layer cannot detect prompt-vocabulary drift for the most common multi-line prose shape.

**Suggested fix:** Scan decoded strings instead of the escaped JSON: serialize to a `JsonNode` (`JsonSerializer.SerializeToNode`), walk all string values, and assert `TrademarkScrubber.ContainsTrademarkedTerm` on each decoded value. Add a guard self-test with a prose field shaped like `"line one\nVDOT line two"` to lock the fix in.

---
*Confidence: 72% | deep-review*

backend/src/RunCoach.Api/Prompts/coaching-system.v1.yaml
```diff
@@ -55,6 +55,10 @@ static_system_prompt: |
   The training paces provided are computed deterministically from the runner's race history using Daniels' Running Formula. You MUST use these exact pace ranges. Do not derive or estimate different paces.
   Volume progression MUST NOT exceed 10% per week. Deload weeks must occur every 3-4 weeks.
 
+  VOCABULARY RULES:
+  - Refer to the pace methodology and the fitness index ONLY as "Daniels-Gilbert zones" or "pace-zone index" — in every prose field you produce (rationales, notes, warnings, week summaries, coaching narrative).
+  - Do NOT use the trademarked four-letter brand name for the Daniels fitness index, in any casing or spelling variant. When you would name the index, say "pace-zone index" instead.
```
leehopper, line 60: **💡 [LOW] Re-recorded safety-boundary eval fixtures have no trademark guard**

**💡 [low] [test_coverage]** Re-recorded safety-boundary eval fixtures have no trademark guard

This PR adds `TrademarkProseGuard.AssertClean` to all five `PlanGenerationEvalTests` methods and to `AdaptationRestructureEvalTests`, but `SafetyBoundaryEvalTests` — the third eval suite whose five `sonnet coachingResponse` fixtures (`safety.medical`, `safety.overtraining`, `safety.injury`, `safety.crisis`, `safety.nutrition`) were **re-recorded in this PR** because the system prompt changed — applies no trademark check to the conversational coaching response.

`GetCoachingResponseAsync` calls `IChatClient` directly, bypassing `ClaudeCoachingLlm` and its scrubber. A re-recorded safety fixture containing the trademarked term would pass the suite silently; the only remaining defense for the conversational surface is the runtime scrub with no eval-time visibility.

**Suggested fix:** In each of the five `SafetyBoundaryEvalTests` scenarios, add one assertion after obtaining `coachingResponse`:
```csharp
TrademarkScrubber.ContainsTrademarkedTerm(coachingResponse).Should().BeFalse(
    because: "safety coaching response must not contain the trademarked pace-index term (Slice 3B F2)");
```
Or wrap in `TrademarkProseGuard.AssertClean("safety-{scenario}", new { coachingResponse })` to match the pattern used in the other eval suites.

---
*Confidence: 67% | deep-review*

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/TrademarkProseGuard.cs
leehopper: Fixed in 383b037. Removed the private `SerializeOptions` field and the `System.Text.Json.Serialization` using directive from `TrademarkProseGuard`; the `AssertClean` path now calls `JsonSerializer.SerializeToNode(output, ClaudeCoachingLlm.StructuredOutputSerializerOptions)` directly.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/TrademarkScrubberTests.cs
leehopper: Fixed in 383b037. Comment now reads `// into Macro.Rationale (Slice 3B F2).` matching the symbolic form used in the XML doc.

backend/src/RunCoach.Api/Modules/Coaching/TrademarkScrubber.cs
leehopper: Fixed in 383b037. `GenerateStructuredAsync` now parses the raw JSON to a `JsonNode`, walks every string leaf with `TrademarkScrubber.ScrubJsonStringValues`, and then deserializes from the scrubbed node — so the regex sees decoded string values (actual newline char before `V`) rather than the raw escaped text (`n` before `V`). A regression test `ScrubJsonStringValues_ReplacesTheTerm_AfterJsonNewlineEscape` covers the `"Key metrics:\nVDOT: 38"` raw-JSON shape that the old path missed.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Eval/TrademarkProseGuard.cs
leehopper: Fixed in 383b037. `AssertClean` now calls `JsonSerializer.SerializeToNode` and walks the `JsonNode` tree with a private `AssertNodeClean` helper that calls `TrademarkScrubber.ContainsTrademarkedTerm` on each decoded string value — same fix as the production path, same root cause eliminated.

backend/src/RunCoach.Api/Prompts/coaching-system.v1.yaml
leehopper: Fixed in 383b037. Added `TrademarkProseGuard.AssertClean("safety-{scenario}", new { coachingResponse })` to all 5 safety scenarios before the verdict assertions. Used the guard pattern rather than a bare `ContainsTrademarkedTerm` call for consistency with the other eval suites.

PR #237, opened 2026-06-30
feat(slice-4b): frontend streaming conversation ux (pr6)

leehopper:
## Slice 4B · PR6 — Frontend streaming conversation UX (Unit 6)

The first frontend PR of the 4B train. Wires the streaming conversation core into the home surface: a hand-rolled SSE reader, live-then-reconcile rendering, the interactive composer, the confirmation card, and the Edit pre-fill. Builds on PR1–PR5 (all merged).

### What's in it

**Streaming foundation**
- **`coach-stream.helpers`** — an incremental SSE decoder (buffers across chunk boundaries, skips the `: hb` heartbeat comment) + a typed frame parser that validates each frame's discriminant primitive fields and drops malformed frames.
- **`coach-stream.model` / `conversation.model`** — the hand-written SSE frame union and the accurate discriminated composed-timeline model (the generated type marks `interactive`/`proactive` both required despite C# nullability).
- **`useCoachStream`** — a hand-rolled `fetch` + `ReadableStream` reader (RTK's `fetchBaseQuery` can't stream). Replicates `baseQueryWith401Handler`: `credentials: 'include'`, the `X-XSRF-TOKEN` double-submit header, the lazy `GET /v1/auth/xsrf` seed on a cold submit, and `401 → loggedOut`. Live tokens render from local React state (never a per-token cache write) and reconcile **exactly once on `done`** via `updateQueryData`. `error` is terminal and the retry re-sends with a **fresh** `clientMessageId`.
- **`conversation.api`** — `getConversationTimeline` query + `confirmConversationalLog` mutation (success-gated `invalidatesTags(['WorkoutLog','Plan','Conversation'])`).
- **`draft-to-form`** — the `StructuredLogDraft → log-form` inverse mapper (km-only collapse, fractional minutes, zero/Skipped → blank).

**UI**
- **`CoachComposer`** — Enter sends, Shift+Enter newlines, send disabled while a stream is in flight.
- **`LogConfirmationCard`** — renders the parsed draft with Confirm / Edit / Cancel.
- **`CoachChat`** — the unified panel: composes the scaffolded `MessageBubble` + aria-live `TranscriptScroller` for interactive turns and reuses `AdaptationTurn` / `SafetyTurn` / `BeforeAfterDiff` for proactive turns from the composed timeline; shows the live streaming bubble, safety notice, and retry affordance; wires Confirm (with failure feedback) / Edit (navigate to the pre-filled log form) / Cancel.
- **`LogPage`** — reads the Edit draft from router state and pre-fills via the inverse mapper.
- **`HomePage`** — mounts `CoachChat` between `TodayCard` and `UpcomingList`, superseding the read-only `ConversationPanel` (the unified timeline includes the proactive turns it showed).

### Scope boundaries (spec non-goals)
- **Plain text only** — `MessageBubble` `whitespace-pre-wrap`, no `react-markdown`, no `dangerouslySetInnerHTML`. Markdown rendering + a sanitizer are deferred behind the research protocol.
- **Tailwind-only animation** with `motion-reduce:` pairing; `motion/react` not adopted this slice.

### Quality
- **TDD throughout** — 42 new unit/component tests + 1 Playwright E2E. Full frontend suite **550 green**, `npm run build` clean, all changed files lint-clean.
- **Adversarial multi-agent review** (5 dimensions → independent verification): 0 findings survived as high/medium. Three genuine low-severity items were fixed — silent confirm-failure (now `reportClientError` + toast, card stays for retry), the SSE parser's unchecked casts (now validated), and a defensive `streamEnded` guard.
- Semantic tokens, `check-contrast`, no `VDOT`.

The Playwright E2E asserts the composer → streamed answer → confirmation card → Confirm flow; it runs on the host stack (`npm run e2e`) since Playwright `route.fulfill` delivers the SSE body in one shot (token-by-token paint is covered by the Vitest hook spec).

## Summary by CodeRabbit

* **New Features**
  * Added interactive coach chat with live streaming responses, safety notices, and log confirmation cards.
  * Added workout-log “Edit” support that pre-fills the form from a confirmed conversational draft (km/duration formatting).
  * Updated the plan home page to display the coach chat between today’s workout and upcoming items.
* **Bug Fixes**
  * Improved robustness of streamed frame decoding/parsing, confirmation handling, and timeline reconciliation, including retry and session-expired behavior.
* **Tests**
  * Expanded end-to-end, API, hook, and component test coverage for streaming, confirmations, retries, and prefill.


leehopper:
@coderabbit review

frontend/src/app/modules/coaching/hooks/use-coach-stream.hooks.ts
```diff
@@ -0,0 +1,377 @@
...
+  const appendTimelineTurn = useCallback(
+    (turn: ConversationTimelineTurnDto): void => {
+      reduxDispatch(
+        conversationApi.util.updateQueryData('getConversationTimeline', undefined, (draft) => {
+          draft.turns.push(turn)
+        }),
+      )
+    },
+    [reduxDispatch],
+  )
```
leehopper, line 304: **💡 [LOW] Cold-start race: optimistic timeline reconcile silently no-ops, dropping the streamed answer**

In the cold-start race where a message is sent before `GET /conversation/timeline` resolves, this `updateQueryData` reconcile is a silent no-op. Verified against the installed RTK 2.12.0 (probe + source): a subscribed-but-pending first-load query has `status: pending` and **no `data` key**, so `updateQueryData` returns without patching — it does **not** throw (it never reaches `draft.turns.push`). On the terminal `done`/`card` frame the user+coach turns then fail to append, the `reconciled` action clears the live bubbles, and — because the hand-rolled streaming endpoint deliberately invalidates no RTK tag — nothing refetches, so the streamed answer vanishes from the transcript with no error shown. The window is narrow (the mount-time GET must still be pending when the terminal frame arrives) but unguarded.

**Suggested fix:** Use `upsertQueryData` (seeds the entry) or guard the recipe so the reconcile survives a still-pending timeline query.

frontend/src/app/modules/coaching/hooks/use-coach-stream.hooks.ts
```diff
@@ -0,0 +1,377 @@
...
+
+  const runStream = useCallback(
+    async (message: string, clientMessageId: string): Promise<void> => {
+      abortRef.current?.abort()
+      const controller = new AbortController()
```
leehopper, line 309: **🟡 [MEDIUM] AbortController lifecycle (re-send abort, unmount abort) is untested**

The `AbortController` lifecycle is untested: `abortRef.current?.abort()` at the start of every send/retry, the unmount effect that aborts on teardown, and the silent `signal.aborted` return branch. The hook spec only sends one message at a time and never unmounts mid-stream — and the existing retry test does not abort a *live* stream (the first stream is already terminal before `retry()` runs, so `abortRef` is null by then). Because each stream optimistically appends the user turn to the cache, a broken abort could double-append or interleave a stale reader.

**Suggested tests:** (1) `send()` then `send()` again before the first completes — assert the first request aborted and no duplicate user turn; (2) `renderHook` then `unmount()` mid-stream — assert the reader aborted and no error state dispatched (silent cancel).

frontend/src/app/modules/coaching/hooks/use-coach-stream.hooks.ts
```diff
@@ -0,0 +1,377 @@
...
+          appendTimelineTurn,
+          coach: { text: '' },
+        })
+      } catch {
+        // An intentional cancel/unmount aborts silently; everything else is a
+        // network failure the user can retry.
+        if (controller.signal.aborted) return
+        dispatch({
+          type: 'error',
+          error: { message: GENERIC_ERROR_MESSAGE, retryable: true, retryAfterSeconds: null },
```
leehopper, line 352: **🟡 [MEDIUM] Non-401 failure-response and network-exception error paths are untested**

Two terminal failure branches here are untested. Only the `401 -> loggedOut` path is exercised; no test drives a 500/503 response, a 200 with a null body (`!response.ok || response.body === null`), or a rejecting `fetch` (the `catch` branch). These are the only paths that surface the user-facing Retry affordance when the coach endpoint is down — exactly the confidently-incomplete error handling the repo AI-code rules call out.

**Suggested tests:** (1) route `/messages` to a 500 -> assert `error.retryable === true`, `isStreaming === false`, and no turn appended to the cache; (2) make `fetch` reject -> assert the same generic retryable error (catch branch).

frontend/src/app/api/conversation.api.ts
```diff
@@ -1,24 +1,58 @@
...
 // antiforgery token (`base-query.ts` only adds X-XSRF-TOKEN to mutations).
 // URL segments are relative to the global `/api` prefix supplied by the base
 // query.
 //
-// `getConversationTurns` is tagged `Conversation` so the `createWorkoutLog`
-// mutation can declare `invalidatesTags: [..., 'Conversation']` and the
-// read-only "Explain-the-change" panel refetches in the same interaction as
-// the plan view (spec 17 § Unit 7).
+// `getConversationTurns` (newest-first proactive turns) feeds the read-only
+// Slice 3 "Explain-the-change" panel; `getConversationTimeline` (oldest-first
```
leehopper: **💡 [LOW] Comment claims getConversationTurns feeds a panel this PR unmounts**

This rewritten comment states, in present tense, that `getConversationTurns` feeds the read-only Slice 3 "Explain-the-change" panel — but this same PR swaps `home.page.tsx` from `ConversationPanel` to `CoachChat`, the only production consumer. `ConversationPanel`, `useConversationTurns`, and `getConversationTurns` are now orphaned (referenced only by their own specs). The comment describes a data-flow that no longer reaches any mounted surface and will mislead a future reader.

**Suggested fix:** Reword to note the Slice 3 panel is no longer mounted, or remove the dead read path if retention is unintended.

leehopper (commented):
### Deep Review — PR #237 (slice-4b frontend streaming conversation UX)

**Frontier-mode multi-agent review** (7 discovery agents + validation + blind challenge). Security, cross-file impact, and code-simplification passes were **clean**; the type-design pass'''s one observation was dismissed at the confidence threshold (by-design, server-trusted draft).

**4 non-blocking findings** posted inline (1 correctness, 2 test-coverage, 1 documentation) — all in the hand-rolled streaming hook or its comment:

- 💡 **bug-1** — cold-start race silently drops a streamed answer (the RTK timeline reconcile no-ops on a still-pending query). Mechanism empirically verified against RTK 2.12.0.
- 🟡 **test-1** — non-401 / network-failure error branches untested.
- 🟡 **test-2** — AbortController lifecycle (re-send abort, unmount abort) untested.
- 💡 **conv-1** — stale comment claims `getConversationTurns` feeds a now-unmounted panel.

No blocking issues — unusually clean for a feature PR of this size.

---
Generated by deep-review | Reviewed up to: 83752d99d1f9d8d45d821acc69d516ee416b788f

frontend/src/app/modules/coaching/components/coach-chat.component.spec.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+  it('renders streamed coach prose as plain text and never mentions VDOT', () => {
+    setTimeline([coachTurn('Run **easy** today, not your pace-zone limit.')])
+    streamMock.mockReturnValue(idleStream())
+
+    renderChat()
+
+    // Markdown is not rendered — the asterisks render literally.
+    expect(screen.getByText(/Run \*\*easy\*\* today/)).toBeInTheDocument()
+    expect(screen.queryByText(/vdot/i)).toBeNull()
```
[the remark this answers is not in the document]
leehopper: Removed the `/vdot/i` sentinel — the negative assertion is itself a frontend trademark-rule violation (it puts the banned term in source). Replaced it with a positive assertion on the approved wording (`/pace-zone limit/`) already rendered by the fixture, and renamed the test accordingly. 8a5e277

frontend/src/app/modules/coaching/components/coach-chat.component.spec.tsx
```diff
@@ -0,0 +1,292 @@
...
+    confirmUnwrap.mockRejectedValue(new Error('confirm boom'))
+    confirmTrigger.mockReturnValue({ unwrap: confirmUnwrap })
+    setTimeline([])
+    streamMock.mockReturnValue(idleStream({ card: sampleCard, dismissCard }))
+
+    renderChat()
+    await user.click(screen.getByRole('button', { name: /^confirm$/i }))
+
+    // The card stays for retry; the user is told it failed and a diagnostic
+    // trail is left (mirrors log.page.tsx's handled-rejection handling).
```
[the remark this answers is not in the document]
leehopper: Reworded to stand on its own — the comment now describes the handled `.unwrap()` rejection behavior inline (otherwise invisible to the global error reporter) without naming another source file. 8a5e277

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (card === null) return
+    try {
+      // A 200 with `adaptation.kind === Error` still means the log committed —
+      // `unwrap` only throws on an HTTP failure, so the card dismisses and the
+      // success-gated invalidation refetches the timeline + plan.
+      await confirmLog({ draft: card.draft, clientMessageId: card.clientMessageId }).unwrap()
+    } catch (error) {
+      // The awaited `.unwrap()` rejection is a *handled* rejection, invisible to
+      // the global error reporter + error boundary, so forward it explicitly and
+      // tell the user — the card stays open for retry (mirrors log.page.tsx).
```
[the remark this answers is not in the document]
leehopper: Dropped the cross-file reference; the comment now explains the handled-rejection / keep-card-open-for-retry behavior inline. 8a5e277

frontend/src/app/modules/coaching/components/log-confirmation-card.component.spec.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    await user.click(screen.getByRole('button', { name: /^cancel$/i }))
+    expect(props.onCancel).toHaveBeenCalledOnce()
+  })
+
+  it('disables Confirm while a commit is in flight and never mentions VDOT', () => {
+    renderCard({ isConfirming: true })
+
+    expect(screen.getByRole('button', { name: /^confirm$/i })).toBeDisabled()
+    expect(screen.queryByText(/vdot/i)).toBeNull()
```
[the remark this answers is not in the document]
leehopper: Dropped the `/vdot/i` negative check (the card renders no zone copy to assert positively) and removed the term from the test name. 8a5e277

frontend/src/app/modules/coaching/hooks/use-coach-stream.hooks.ts
leehopper: Confirmed and fixed. Verified against the installed RTK 2.12.0 (source + a probe): on a cold cache `updateQueryData` produces zero patches (no `data` to patch, no throw), and `upsertQueryData` is **not** a safe alternative — the later-resolving mount GET keeps the original requestId and clobbers the seeded turns.

Fix in `appendTimelineTurn`: capture the `PatchCollection` and, when `patches.length === 0`, dispatch `invalidateTags(['Conversation'])`. RTK's default *delayed* invalidation queues the refetch behind the in-flight mount GET and pulls the server-persisted turns once it settles; the warm path patches in place (non-empty patches) and skips the fallback, so there's no extra fetch. New hook test: `reconciles via cache invalidation when the optimistic push lands on a cold timeline`. 8a5e277

frontend/src/app/modules/coaching/hooks/use-coach-stream.hooks.ts
leehopper: Added the two suggested hook tests:
1. `send()` then `send()` again before the first terminates — asserts the first request's signal aborted and that only the second user turn reconciles (no duplicate user turn).
2. `unmount()` mid-stream — asserts the controller aborts and the reader cancels silently (no error surfaced, no turn appended).

A `controller.error(...)` hook was added to the controlled-stream helper to simulate the reader rejection a signal-honouring fetch raises on abort (the stubbed fetch ignores the signal). 8a5e277

frontend/src/app/modules/coaching/hooks/use-coach-stream.hooks.ts
leehopper: Added three hook tests for the previously-untested terminal branches:
1. a 500 response (`!response.ok`) → `error.retryable === true`, `isStreaming === false`, no turn appended, and auth status unchanged (not logged out).
2. a 200 with a null body (`response.body === null`) → same generic retryable error.
3. a rejecting `fetch` (the `catch` branch) → same generic retryable error, no turn appended.
8a5e277

frontend/src/app/api/conversation.api.ts
leehopper: Reworded. The comment no longer claims `getConversationTurns` feeds a mounted panel — it now notes the turns feed has no mounted consumer (the home route renders the timeline-backed chat) while its query / hook / tag wiring is retained for reuse.

Confirmed orphaned: `getConversationTurns` → `useGetConversationTurnsQuery` → `useConversationTurns`, plus `ConversationPanel`, are referenced only by their own specs — no mounted production surface. Full dead-code removal (endpoint + hook + panel + their tests) is a larger cleanup left as a follow-up rather than folded into this review pass. 8a5e277

PR #255, opened 2026-07-03
feat(slice-4c-units): PR6 — miles→km conversion at the log write site

leehopper:
## Slice 4C-units — PR6 (final PR of the sub-slice)

Makes the one real numeric **write** unit-aware: a Miles-preferring runner enters distance in miles on the `/log` form and it persists as canonical km/SI, while storage, the wire, and the plan-gen prompt stay km-native and **the LLM performs zero conversion** (DEC-086). Frontend-only.

### What changed
- **Shared module** (`unit-format.helpers.ts`): the single write conversion `preferredDistanceToMeters` (km path byte-identical to the prior `distanceKm * 1000`), its inverse `metresToPreferredDistance` (draft pre-fill), and `distanceUnitLabel`. All unit math lives in one module — one shared conversion, one write site, one `1609.344`.
- **Schema** (`workout-log-form.schema.ts`): the form-local distance field is renamed **`distanceKm` → `distance`** because it now holds the value in the runner's *display* unit (km OR miles) — `distanceKm` would be a misnomer under miles. The schema is now a `makeWorkoutLogFormSchema(units)` factory: unit-aware label + a `Math.ceil` mile-equivalent `max` (622 mi — a client sanity guard that is only ever *looser* than the km-native 1000 km ceiling, never stricter). `toCreateWorkoutLogRequest` takes a **required** `units` param.
- **`/log` page**: split into an outer that resolves the preference (new `usePreferredUnitsResolution`) and gates, and an inner form that mounts only once the unit is **known**. This means the miles→km write is never interpreted against the loading-time km fallback (a cold-cache data-corruption risk), the draft-edit pre-fill computes once in the correct unit, and there's no mid-form schema swap.
- **Draft pre-fill** (`draft-to-form.helpers.ts`): units-aware, and short-circuits the identity case (draft already in the preferred unit) to avoid FP round-trip drift (`100 mi → 99.99999999999999`).
- The **log-confirmation card is deliberately untouched** — it keeps echoing the runner-*stated* unit (build decision, per spec).

### Rename footprint
The `distanceKm → distance` rename is deliberate (the honest structural fix, not scope creep): it touches 3 prod + 4 spec files but the value at the conversion site now reads `values.distance * METERS_PER_MILE` instead of a `distanceKm`-named value multiplied by metres-per-mile.

### Design de-risking
The design was **adversarially red-teamed before implementation** (3 independent lenses + synthesis). Top finding — feeding the display-only km fallback into the *write* path could silently persist km for a miles runner on a cold cache — drove the resolve-then-mount split above. FP round-trip drift, the max-bound rounding direction (`ceil`, not `round`), and the resolver-swap/revalidation risk were all folded in.

### Proofs
- Schema unit test: a miles-entered distance round-trips to the correct `distanceMeters` (5 mi → `5 * 1609.344`). **[T06.2]**
- 109 unit/RTL tests across the five touched specs; full frontend suite **642/642**.
- Coaching **eval suite 193/193 green in Replay** locally (`--filter-namespace …Coaching.Eval*`); zero backend/prompt/wire files changed, so the km-native prompt and wire are unchanged by construction. **[T06.3, acceptance criterion 4]**
- `npm run build` (tsc + vite) green; touched files ESLint-clean.

### Follow-ups found
- No Miles-path E2E yet (`e2e/workout-logging.spec.ts` covers km only; it relies on fresh accounts server-side-defaulting to Kilometers). Tracked for the slice-level live pass.
- Onboarding weekly-distance entry is a second, LLM-mediated distance write path — explicitly out of scope here (belongs to `4C-onboarding`).

## Summary by CodeRabbit

* **New Features**
  * Logging now supports preferred distance units, showing the distance input label and values in **km** or **mi** based on user settings.
  * The **/log** page waits for unit preferences to resolve before rendering the form.

* **Bug Fixes**
  * Miles and kilometers are now converted and saved consistently, including edit/draft pre-fill without round-trip drift.
  * Empty distance and duration fields remain blank (no unwanted `"0"` values).

* **Tests**
  * Updated and expanded unit-conversion and form-schema coverage.

### DIFF
--- a/frontend/src/app/modules/logging/pages/log.page.tsx
+++ b/frontend/src/app/modules/logging/pages/log.page.tsx
@@ -4,6 +4,7 @@ import { useLocation, useNavigate } from 'react-router-dom'
 import { zodResolver } from '@hookform/resolvers/zod'
 import { toast } from 'sonner'
 
+import { Button } from '@/components/ui/button'
 import { PreferredUnits, type StructuredLogDraft } from '~/api/generated'
 import { useCreateWorkoutLogMutation } from '~/api/workout-log.api'
 import { reportClientError } from '~/error-boundary/report-client-error'
@@ -107,18 +108,85 @@ const WorkoutLogFormView = ({ units, editDraft }: WorkoutLogFormViewProps) => {
   )
 }
 
+interface UnitPreferenceUnavailableProps {
+  onRetry: () => void
+}
+
+/**
+ * Shown when the unit-preference query errors. The form stays gated rather than
+ * falling back to km, because `GET /settings/units` and the log `POST` are
+ * independent endpoints — silently defaulting to km could persist a miles
+ * runner's distance at ~1.6× the wrong magnitude (DEC-086). Retry re-runs the
+ * preference query.
+ */
+const UnitPreferenceUnavailable = ({ onRetry }: UnitPreferenceUnavailableProps) => (
+  <div
+    role="alert"
+    data-testid="log-page-units-error"
+    className="flex flex-col items-center gap-3 py-8 text-center"
+  >
+    <p className="text-sm text-muted-foreground">
+      We couldn’t load your distance-unit preference. Check your connection and try again.
+    </p>
+    <Button type="button" size="sm" onClick={onRetry}>
+      Retry
+    </Button>
+  </div>
+)
+
+interface LogPageContentProps {
+  units: PreferredUnits
+  isResolved: boolean
+  isError: boolean
+  refetch: () => void
+  editDraft: StructuredLogDraft | null
+  /** `location.key` — remounts the form on each fresh /log navigation (see below). */
+  formKey: string
+}
+
+/**
+ * Selects what renders below the `/log` header based on the preference query
+ * state: a retry on error, a loading placeholder until settled, otherwise the
+ * form. Split out of {@link LogPage} so the three states read as early returns
+ * rather than a nested ternary.
+ */
+const LogPageContent = ({
+  units,
+  isResolved,
+  isError,
+  refetch,
+  editDraft,
+  formKey,
+}: LogPageContentProps) => {
+  if (isError) {
+    return <UnitPreferenceUnavailable onRetry={refetch} />
+  }
+  if (!isResolved) {
+    return (
+      <p role="status" className="text-sm text-muted-foreground" data-testid="log-page-loading">
+        Loading…
+      </p>
+    )
+  }
+  // Key on `location.key` so re-navigating to /log with a different draft remounts
+  // the form: `useForm` only seeds `defaultValues` on mount, so a new editDraft
+  // would otherwise leave the previous draft's fields in place.
+  return <WorkoutLogFormView key={formKey} units={units} editDraft={editDraft} />
+}
+
 /**
  * The `/log` route — a mobile-first workout-logging form. The form's numeric
  * write interprets the entered distance in the runner's unit preference, so it
  * defers the form until that preference has resolved (usually instant — the
  * preference is cached app-wide; only a cold `/log` refresh shows the brief
  * loading state). This keeps a miles-preferring runner's input from ever being
- * converted against the loading-time km fallback (DEC-086).
+ * converted against the loading-time km fallback (DEC-086); an errored read
+ * surfaces a retry rather than falling through to km.
  */
 const LogPage = () => {
   const location = useLocation()
   const editDraft = (location.state as LogPageLocationState | null)?.draft ?? null
-  const { units, isResolved } = usePreferredUnitsResolution()
+  const { units, isResolved, isError, refetch } = usePreferredUnitsResolution()
 
   return (
     <main
@@ -129,13 +197,14 @@ const LogPage = () => {
         <h1 className="text-2xl font-semibold text-foreground">Log a workout</h1>
         <p className="text-sm text-muted-foreground">Record what you actually ran.</p>
       </header>
-      {isResolved ? (
-        <WorkoutLogFormView units={units} editDraft={editDraft} />
-      ) : (
-        <p role="status" className="text-sm text-muted-foreground" data-testid="log-page-loading">
-          Loading…
-        </p>
-      )}
+      <LogPageContent
+        units={units}
+        isResolved={isResolved}
+        isError={isError}
+        refetch={refetch}
+        editDraft={editDraft}
+        formKey={location.key}
+      />
     </main>
   )
 }
--- a/frontend/src/app/modules/logging/pages/log.page.tsx
+++ b/frontend/src/app/modules/logging/pages/log.page.tsx
@@ -4,56 +4,62 @@ import { useLocation, useNavigate } from 'react-router-dom'
 import { zodResolver } from '@hookform/resolvers/zod'
 import { toast } from 'sonner'
 
-import type { StructuredLogDraft } from '~/api/generated'
+import { PreferredUnits, type StructuredLogDraft } from '~/api/generated'
 import { useCreateWorkoutLogMutation } from '~/api/workout-log.api'
 import { reportClientError } from '~/error-boundary/report-client-error'
+import { distanceUnitLabel } from '~/modules/common/utils/unit-format.helpers'
 import { LogForm } from '~/modules/logging/components/log-form.component'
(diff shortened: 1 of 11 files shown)


leehopper (commented):
### 🤖 Deep Review — PR #255

Frontier-mode review (7 Opus discovery agents → deterministic verification → Sonnet validation → blind Opus challenge). **3 findings**, all on `use-preferred-units.hooks.ts`: one correctness gap (💡 low) plus two improvement suggestions (hook naming + comment accuracy). No blocking issues — the conversion math, the `distanceKm→distance` rename, the schema-factory refactor, and every cross-file signature update verified clean, and 5 of 7 dimensions came back empty. See inline comments.

frontend/src/app/modules/settings/hooks/use-preferred-units.hooks.ts
```diff
@@ -10,8 +10,43 @@ import { useGetUnitPreferenceQuery } from '~/api/settings.api'
...
+ * the {@link PreferredUnits.Kilometers} default is authoritative (the `GET` is
+ * read-or-default 200, so an error means the whole API — including the log `POST`
+ * — is unavailable, and no wrong-unit write can be persisted anyway).
+ */
+export const usePreferredUnitsResolution = (): PreferredUnitsResolution => {
+  const { data, isLoading } = useGetUnitPreferenceQuery(undefined)
+  return {
+    units: data?.preferredUnits ?? PreferredUnits.Kilometers,
+    isResolved: !isLoading,
+  }
```
leehopper: **💡 [LOW] Errored settings GET silently converts a Miles-preference distance as km on write**

**💡 [low] [bug] Errored settings GET silently converts a Miles-preference distance as km on write**

RTK Query’s `isLoading` goes false on **both** success and error, so `isResolved: !isLoading` treats an **errored** `GET /settings/units` as resolved: `isResolved` flips true while `units` falls back to `Kilometers`, and the `/log` form mounts and accepts a submit. `GET /settings/units` and `POST .../workouts/logs` are independent endpoints, and the base query has no retry wrapper / no `refetchOnMountOrArgChange` (the errored GET is cached). So a Miles-preferring runner whose settings GET failed transiently but whose POST succeeds has their `5` (meaning 5 mi) persisted as **5000 m instead of 8046.72 m** — a silent ~1.6× wrong-unit write, the exact case the resolution gate exists to prevent. (`usePreferredUnits` also fires app-wide on home/coach-chat/history, so a stale boot-time error can carry into `/log`.)

**Suggested fix:** Distinguish error from success — read `isError`/`isSuccess` from the query and treat an errored settings read as *not resolved* (keep the gate up, or surface an explicit retry/error state for the write path) rather than silently defaulting to km and allowing the write. Don’t rely on the assumption that a settings-endpoint failure implies the log-POST endpoint is also down.

---
*Confidence: 80% | deep-review*

frontend/src/app/modules/settings/hooks/use-preferred-units.hooks.ts
```diff
@@ -10,8 +10,43 @@ import { useGetUnitPreferenceQuery } from '~/api/settings.api'
...
+export interface PreferredUnitsResolution {
+  /** The resolved preference, or the {@link PreferredUnits.Kilometers} default once settled. */
+  units: PreferredUnits
+  /**
+   * `false` until the preference query has settled (success OR error). The write
+   * path must wait for this before interpreting a typed distance, so a miles
+   * runner's input is never converted against the loading-time km fallback.
+   */
+  isResolved: boolean
+}
```
leehopper: **💡 [LOW] Hook return interface not named Use{HookName}Return**

**💡 [low] [convention] New hook return interface not named `Use{HookName}Return`**

The hook’s return interface is named `PreferredUnitsResolution` rather than `UsePreferredUnitsResolutionReturn`. `frontend/CLAUDE.md` § Custom Hooks mandates `Use{HookName}Return`, and the repo follows it everywhere else (`UseAuthReturn`, `UsePlanReturn`, `UseCoachStreamReturn`, `UseConversationTurnsReturn`, `UseOnboardingTurnReturn`) with no counter-examples.

**Suggested fix:** Rename to `UsePreferredUnitsResolutionReturn` and update the import + annotation in `log.page.tsx` and `log.page.spec.tsx`.

---
*Confidence: 65% | deep-review*

frontend/src/app/modules/settings/hooks/use-preferred-units.hooks.ts
```diff
@@ -10,8 +10,43 @@ import { useGetUnitPreferenceQuery } from '~/api/settings.api'
...
+
+/**
+ * Reads the unit preference for the **write** path — pairs the resolved unit with
+ * a settled flag so a numeric write (the log form) can hold off until the runner's
+ * unit is actually known, rather than converting against the display fallback.
+ *
+ * `isResolved` flips true on both success and error: once the query has settled,
+ * the {@link PreferredUnits.Kilometers} default is authoritative (the `GET` is
+ * read-or-default 200, so an error means the whole API — including the log `POST`
+ * — is unavailable, and no wrong-unit write can be persisted anyway).
```
leehopper: **💡 [LOW] Doc comment overstates a safety invariant about an errored settings GET**

**💡 [low] [comment-accuracy] Doc comment overstates a safety invariant about an errored settings GET**

This comment asserts an errored settings GET means “the whole API — including the log `POST` — is unavailable, and no wrong-unit write can be persisted anyway.” That is an absolute guarantee the code does not provide: `GET /settings/units` and the log `POST` are independent endpoints that can fail separately, so on an isolated GET error the form still renders with the km default and a Miles runner can submit at the wrong magnitude. This is the documentation face of the correctness finding on lines 46–51.

**Suggested fix:** Soften the claim to a best-effort km fallback on error, or make the error path explicit and update the comment to match.

---
*Confidence: 60% | deep-review*

frontend/src/app/modules/logging/pages/log.page.tsx
```diff
@@ -4,56 +4,62 @@ import { useLocation, useNavigate } from 'react-router-dom'
...
+  const schema = useMemo(() => makeWorkoutLogFormSchema(units), [units])
   const defaultValues = useMemo(
     () =>
       editDraft !== null
-        ? draftToWorkoutLogFormFields(editDraft)
+        ? draftToWorkoutLogFormFields(editDraft, units)
         : makeDefaultWorkoutLogFormFields(),
-    [editDraft],
+    [editDraft, units],
   )
```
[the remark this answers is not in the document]
leehopper: Fixed in 03c39a5 — `WorkoutLogFormView` is now keyed off `location.key` in `LogPage`, so re-navigating to `/log` with a different `state.draft` remounts the form and `useForm` re-seeds `defaultValues` from the new draft. The three render states (error / loading / form) were extracted into `LogPageContent` so this reads as an early return rather than a nested ternary.

frontend/src/app/modules/settings/hooks/use-preferred-units.hooks.ts
leehopper: Fixed in 03c39a5. `usePreferredUnitsResolution` now gates `isResolved` on `isSuccess` (not `!isLoading`) and additionally returns `isError` + `refetch`. An errored `GET /settings/units` no longer resolves to the km default — `/log` renders a retry affordance (`UnitPreferenceUnavailable`) instead of the form, so a Miles-preferring runner can never submit against the km fallback after a transient settings-read failure. New tests cover the errored path in both `use-preferred-units.hooks.spec.ts` and `log.page.spec.tsx`.

frontend/src/app/modules/settings/hooks/use-preferred-units.hooks.ts
leehopper: Fixed in 03c39a5 — renamed `PreferredUnitsResolution` → `UsePreferredUnitsResolutionReturn` per the `Use{HookName}Return` convention, and updated the import + type annotations in `log.page.tsx` and `log.page.spec.tsx`.

frontend/src/app/modules/settings/hooks/use-preferred-units.hooks.ts
leehopper: Fixed in 03c39a5 — the doc comment no longer asserts that an errored settings GET implies the whole API (including the log `POST`) is unavailable. It now states the errored read is reported via `isError` and that callers gate the write on `isResolved` and offer a retry on error, matching the corrected behavior in the finding above.

PR #263, opened 2026-07-06
feat(slice-4c-onboarding): deterministic form-answer origination endpoint

leehopper:
## Slice 4C-onboarding · PR-C (DU-1) — deterministic form-answer origination endpoint

Net-new **additive** `POST /api/v1/onboarding/answers`: a deterministic, form-first onboarding intake (DEC-086 D1) that originates the **existing** `AnswerCaptured` event per submitted topic with **no LLM call**, reusing the event-sourced spine — both projections, the completion gate, and the inline plan-generation terminal branch — byte-for-byte unchanged (DEC-047/DEC-060 hold). It **coexists** with the legacy conversational `/turns` path (retired later in PR-E).

### What's here

- **Command + plain static Wolverine handler** (`SubmitStructuredAnswers` / `SubmitStructuredAnswersHandler`, DP-2 — mirrors `OnboardingTurnHandler`, not `[AggregateHandler]`): idempotency short-circuit → `StartStreamOrAppendAsync` bootstrap (the shared per-user stream, PR-A #259) → one whole-record `AnswerCaptured` per topic (`Confidence 1.0`, **default PascalCase** for projection round-trip parity) → completion gate as the **sole** plan-gen authority (no LLM `ReadyForPlan`) → inline `GeneratePlanAsync` + `PlanLinkedToUser` + `OnboardingCompleted`. No `SaveChangesAsync` — Wolverine brackets the transaction.
- **6 loosened wire-input DTOs + `SubmitStructuredAnswersRequestMapper`** (FR-1.8, the deterministic replacement for the retired LLM validator). Rationale: the canonical `*Answer` records self-validate in their `init` accessors and throw `TargetInvocationException` if bound directly (→ uncatchable **500**). So the wire DTOs are loosened primitives that bind without throwing; the mapper validates deterministically (enum definedness, ISO-8601 date/duration, **finite + ranged** distances, the `TargetEvent ⇒ RaceTraining` cross-field rule) and constructs the canonical records in `try/catch (ArgumentException)` → clean **400** — the repo idiom (`WorkoutLogsController`).
- **Controller `SubmitAnswers`**: validate → dispatch → reload committed view (mirrors `ReviseAnswer`) → `200` / `400` / `409` (already complete) / `422` (plan-gen rejected — nothing staged, re-submittable).
- **Regenerated OpenAPI + rtk/zod** for PR-D.

### Verification

- 32 mapper unit tests + 12 integration tests (both projections materialize from the form-originated event; fitness/partial/partial-then-complete merge; duplicate-key no-double-append; shared-stream bootstrap collision; 409/422; every FR-1.8 hostile-input path → 400, incl. an overflow numeric literal `1e400` → **400 not 500**).
- Full backend suite **green in Replay**. Release build passes the SonarAnalyzer/StyleCop gate. Frontend build + 627 tests green.

### Review (adversarial, blind-challenge verified) — 9 findings confirmed, all addressed

| # | Sev | Finding | Resolution |
|---|-----|---------|-----------|
| 3 | **critical** | `1e400` → `+Infinity` slips past the records' one-sided guards, then crashes `SerializeToDocument` as a 500 | **Fixed** — mapper rejects non-finite / absurd distances; end-to-end test asserts 400 |
| 4/5 | med/nit | `XmlConvert.ToTimeSpan` accepts negative / absurd durations | **Fixed** — reject `< 0` / `> 60d` |
| 2/6 | low/nit | cross-field check read the enum before `Enum.IsDefined` → misleading (still 400) message | **Fixed** — validate goal enum first |
| 7 | high | no shared-stream bootstrap-collision test | **Added** |
| 8 | high | no partial-then-complete (save-and-resume) test | **Added** |
| 9 | med | all-six mapper test spot-checked 3 topics | **Fixed** — full-record equality for all 6 |
| 1 | med | **double-submit with *different* idempotency keys** can double-generate a plan + leave `CurrentPlanId` nondeterministic | **Deferred + tracked** (see below) |

### Deferred, tracked follow-ups (builder decision)

- **Finding #1 (concurrency).** The plain static handler has no Marten optimistic-concurrency gate, so two concurrent submissions with *different* keys both pass the gate and both generate a plan. This is **inherited verbatim from `OnboardingTurnHandler`** (DP-2 deliberately mirrored that unprotected convention over `[AggregateHandler]`/`FetchForWriting`). Fixing only the new handler would diverge from DP-2 and from the shipped turn handler → this needs a **cross-cutting decision applied to both onboarding handlers** (add expected-version optimistic concurrency + `ConcurrencyException → 409`), tracked for the builder. Practical mitigation lands in PR-D (disable the submit button on click); MVP-0 audience is self+family.
- **Generated request DTO marks all topic slots `required`** despite being C#-nullable — the known repo-wide `RequireNonNullablePropertiesSchemaFilter` `$ref` artifact (deferred structural fix, PR #163). Runtime binding correctly treats topics as optional; PR-D hand-writes accurate optional-topic models like `plan.model.ts`.

Pre-existing frontend lint noise (8 `sonarjs/no-hardcoded-passwords` on password/e2e fixtures) is untouched by this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a new structured onboarding answers API that accepts standardized inputs for current fitness, injury history, preferences, primary goal, target event, and weekly schedule.
  * Submissions now support idempotent handling and return deterministic outcomes, including whether a plan was generated.
* **Bug Fixes**
  * Strengthened request validation (required fields, numeric/date/duration formats, and cross-field consistency) with clean 400/409/422 responses.
  * Prevents duplicate staging and additional plans on repeated submissions with the same idempotency key.
* **Tests**
  * Added end-to-end integration coverage and extensive request-mapper unit tests for valid and hostile inputs.

### DIFF
--- a/backend/openapi/swagger.json
+++ b/backend/openapi/swagger.json
@@ -894,6 +894,122 @@
         ]
       }
     },
+    "/api/v1/onboarding/answers": {
+      "post": {
+        "tags": [
+          "Onboarding"
+        ],
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/SubmitStructuredAnswersRequestDto"
+              }
+            },
+            "text/json": {
+              "schema": {
+                "$ref": "#/components/schemas/SubmitStructuredAnswersRequestDto"
+              }
+            },
+            "application/*+json": {
+              "schema": {
+                "$ref": "#/components/schemas/SubmitStructuredAnswersRequestDto"
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "$ref": "#/components/schemas/OnboardingStateDto"
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/OnboardingStateDto"
+                }
+              },
+              "text/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/OnboardingStateDto"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Request",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              },
+              "text/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              }
+            }
+          },
+          "409": {
+            "description": "Conflict",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              },
+              "text/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              }
+            }
+          },
+          "422": {
+            "description": "Unprocessable Content",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              },
+              "text/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              }
+            }
+          }
+        },
+        "security": [
+          {
+            "cookieAuth": [ ]
+          },
+          {
+            "bearerAuth": [ ]
+          }
+        ]
+      }
+    },
     "/api/v1/plan/current": {
       "get": {
         "tags": [
@@ -1846,6 +1962,37 @@
         },
         "additionalProperties": false
       },
+      "CurrentFitnessInputDto": {
+        "required": [
+          "longestRecentRunKm",
+          "typicalWeeklyKm"
+        ],
+        "type": "object",
+        "properties": {
+          "typicalWeeklyKm": {
+            "type": "number",
+            "format": "double"
+          },
+          "longestRecentRunKm": {
+            "type": "number",
+            "format": "double"
+          },
+          "recentRaceDistanceKm": {
+            "type": "number",
+            "format": "double",
+            "nullable": true
+          },
+          "recentRaceTimeIso": {
+            "type": "string",
+            "nullable": true
+          },
+          "description": {
+            "type": "string",
+            "nullable": true
+          }
+        },
+        "additionalProperties": false
+      },
       "DaySlotType": {
         "enum": [
           "Run",
@@ -1884,6 +2031,26 @@
           "pastInjurySummary": {
             "type": "string",
             "description": "Summary of past injuries or recurring issues that should inform the training plan. Empty string if none."
+          }
+        },
+        "additionalProperties": false
+      },
+      "InjuryHistoryInputDto": {
+        "required": [
+          "hasActiveInjury"
+        ],
+        "type": "object",
+        "properties": {
+          "hasActiveInjury": {
+            "type": "boolean"
+          },
+          "activeInjuryDescription": {
+            "type": "string",
+            "nullable": true
+          },
+          "pastInjurySummary": {
+            "type": "string",
+            "nullable": true
           }
         },
         "additionalProperties": false
@@ -2420,6 +2587,30 @@
         },
         "additionalProperties": false
       },
+      "PreferencesInputDto": {
+        "required": [
+          "comfortableWithIntensity",
+          "preferredUnits",
+          "preferTrail"
+        ],
+        "type": "object",
+        "properties": {
+          "preferredUnits": {
+            "$ref": "#/components/schemas/PreferredUnits"
+          },
+          "preferTrail": {
+            "type": "boolean"
+          },
+          "comfortableWithIntensity": {
+            "type": "boolean"
+          },
+          "description": {
+            "type": "string",
+            "nullable": true
+          }
+        },
+        "additionalProperties": false
+      },
       "PreferredUnits": {
         "enum": [
           0,
@@ -2452,6 +2643,22 @@
           "description": {
             "type": "string",
             "description": "Runner-supplied free-text description that informed the categorical mapping."
+          }
+        },
+        "additionalProperties": false
+      },
+      "PrimaryGoalInputDto": {
+        "required": [
+          "goal"
+        ],
+        "type": "object",
+        "properties": {
+          "goal": {
+            "$ref": "#/components/schemas/PrimaryGoal"
+          },
+          "description": {
+            "type": "string",
+            "nullable": true
           }
         },
         "additionalProperties": false
@@ -2687,6 +2894,43 @@
         },
         "additionalProperties": false
       },
+      "SubmitStructuredAnswersRequestDto": {
+        "required": [
+          "currentFitness",
+          "idempotencyKey",
+          "injuryHistory",
+          "preferences",
+          "primaryGoal",
+          "targetEvent",
+          "weeklySchedule"
+        ],
+        "type": "object",
+        "properties": {
+          "idempotencyKey": {
+            "type": "string",
+            "format": "uuid"
+          },
+          "primaryGoal": {
+            "$ref": "#/components/schemas/PrimaryGoalInputDto"
+          },
+          "targetEvent": {
+            "$ref": "#/components/schemas/TargetEventInputDto"
+          },
+          "currentFitness": {
+            "$ref": "#/components/schemas/CurrentFitnessInputDto"
+          },
+          "weeklySchedule": {
+            "$ref": "#/components/schemas/WeeklyScheduleInputDto"
+          },
+          "injuryHistory": {
+            "$ref": "#/components/schemas/InjuryHistoryInputDto"
+          },
+          "preferences": {
+            "$ref": "#/components/schemas/PreferencesInputDto"
+          }
+        },
+        "additionalProperties": false
+      },
       "SuggestedInputType": {
         "enum": [
           0,
@@ -2723,6 +2967,31 @@
           "targetFinishTimeIso": {
             "type": "string",
             "description": "Optional target finishing time as ISO-8601 duration (e.g. PT1H45M30S). Null if the runner has no specific time goal.",
+            "nullable": true
+          }
+        },
+        "additionalProperties": false
+      },
+      "TargetEventInputDto": {
+        "required": [
+          "distanceKm",
+          "eventDateIso",
+          "eventName"
+        ],
+        "type": "object",
+        "properties": {
+          "eventName": {
+            "type": "string"
+          },
+          "distanceKm": {
+            "type": "number",
+            "format": "double"
+          },
+          "eventDateIso": {
+            "type": "string"
+          },
+          "targetFinishTimeIso": {
+            "type": "string",
             "nullable": true
           }
         },
@@ -2835,6 +3104,56 @@
           "description": {
             "type": "string",
             "description": "Runner-supplied free-text description of constraints not captured by the day flags (e.g. 'no early mornings')."
+          }
+        },
+        "additionalProperties": false
+      },
+      "WeeklyScheduleInputDto": {
+        "required": [
+          "friday",
+          "maxRunDaysPerWeek",
+          "monday",
+          "saturday",
+          "sunday",
+          "thursday",
+          "tuesday",
+          "typicalSessionMinutes",
+          "wednesday"
+        ],
+        "type": "object",
+        "properties": {
+          "maxRunDaysPerWeek": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "typicalSessionMinutes": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "monday": {
+            "type": "boolean"
+          },
+          "tuesday": {
+            "type": "boolean"
+          },
+          "wednesday": {
+            "type": "boolean"
+          },
+          "thursday": {
+            "type": "boolean"
+          },
+          "friday": {
+            "type": "boolean"
+          },
+          "saturday": {
+            "type": "boolean"
+          },
+          "sunday": {
+            "type": "boolean"
+          },
+          "description": {
+            "type": "string",
+            "nullable": true
           }
         },
         "additionalProperties": false
--- a/backend/src/RunCoach.Api/Modules/Coaching/Onboarding/Models/SubmitStructuredAnswersRequestDto.cs
+++ b/backend/src/RunCoach.Api/Modules/Coaching/Onboarding/Models/SubmitStructuredAnswersRequestDto.cs
@@ -1,0 +1,28 @@
+using System.Text.Json.Serialization;
+
+namespace RunCoach.Api.Modules.Coaching.Onboarding.Models;
+
+/// <summary>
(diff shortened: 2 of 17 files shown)


leehopper (commented):
## Deep Review — PR #263 (Frontier mode)

7 discovery agents → verify → 2 validators → 6 Opus blind-challengers. 1 real defect + 4 low-severity notes. Only the first is a genuine bug I'd recommend fixing before merge; the rest are optional.

_Note: 3 findings (incl. the bug below) were initially eliminated by the verifier because agents cited diff line numbers instead of file line numbers — re-anchored and re-verified so they got fair challenge rather than being silently dropped._

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/SubmitStructuredAnswersRequestMapper.cs
```diff
@@ -0,0 +1,317 @@
...
+            if (duration < TimeSpan.Zero || duration > MaxDuration)
+            {
+                normalized = null;
+                return false;
+            }
+
+            normalized = value;
+            return true;
+        }
+        catch (FormatException)
```
leehopper: **🟡 [MEDIUM] OverflowException from XmlConvert.ToTimeSpan escapes to an unhandled 500**

`XmlConvert.ToTimeSpan` does **checked** arithmetic and throws `System.OverflowException` (an `ArithmeticException`, **not** `FormatException`) for a syntactically valid but numerically huge `xsd:duration` such as `"P1000000000D"` (1e9 days). This method catches only `FormatException`; the caller `TryMap` catches only `ArgumentException`; and `OnboardingController.SubmitAnswers` calls `TryMap` **outside** its `try` block. So the exception propagates unhandled → **HTTP 500**.

This is the duration-field twin of the `1e400`→`+Infinity` distance hole you already closed — the exact failure mode FR-1.8 exists to prevent. Both reachable fields (`TargetEvent.TargetFinishTimeIso`, `CurrentFitness.RecentRaceTimeIso`) hit it, and it's untested (the suite only exercises `P100Y`, which stays in range and is correctly rejected by the `MaxDuration` check).

Suggest widening the catch and adding a test asserting `"P1000000000D"` → 400.

```suggestion
        catch (Exception ex) when (ex is FormatException or OverflowException)
```

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/OnboardingController.cs
```diff
@@ -208,6 +211,88 @@ public async Task<IActionResult> ReviseAnswer(
         return Ok(BuildStateDto(view));
     }
 
+    /// <summary>POST /api/v1/onboarding/answers — submit structured onboarding answers (form-first intake).</summary>
```
leehopper, line 215: **💡 [LOW] Class-summary XML doc omits the new SubmitAnswers endpoint**

The class-level `<summary>` (top of `OnboardingController`) still enumerates three endpoints — "submit a turn, read current state, revise an answer." This PR adds a fourth public action (`SubmitAnswers`, `POST /api/v1/onboarding/answers`) to the same partial class, so that enumeration is now stale. Consider adding this endpoint to the class summary.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/OnboardingController.cs
```diff
@@ -208,6 +211,88 @@ public async Task<IActionResult> ReviseAnswer(
...
+                title: "We couldn't build a plan from your answers.",
+                detail: "Please submit again.",
+                statusCode: StatusCodes.Status422UnprocessableEntity);
+        }
+
+        // Reload the committed projection so the response reflects authoritative state (mirrors
+        // ReviseAnswer). The handler always bootstraps + appends, so the view must exist post-commit.
+        await using var session = store.LightweightSession(userId.ToString());
+        var view = await session.LoadAsync<OnboardingView>(userId, ct).ConfigureAwait(false);
+        if (view is null)
```
leehopper, line 285: **💡 [LOW] StateUnavailable 500 branch has zero test coverage**

The post-commit reload's `view is null` → 500 branch is never exercised by any test (all coverage is via `WebApplicationFactory` asserting 200/400/409/422; no controller-level unit harness exists). This is provably-unreachable defensive code given the handler's always-bootstrap-then-append contract, so this is low value — noting it only for completeness.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/SubmitStructuredAnswersHandler.cs
```diff
@@ -0,0 +1,242 @@
...
+            throw new OnboardingAlreadyCompleteException(cmd.UserId);
+        }
+
+        // Working copy so the completion gate sees the in-flight state without waiting for Marten to
+        // materialize the projection mid-handler.
+        var working = view ?? CreateBootstrapView(streamId, now);
+
+        var topicsCaptured = 0;
+
+        if (cmd.PrimaryGoal is not null)
```
leehopper, line 114: **💡 [LOW] Six topic blocks share an identical append+clear+increment tail**

The six `if (cmd.X is not null)` blocks repeat an identical `AppendAnswer(...) → ClearClarification(...) → topicsCaptured++` tail, varying only the topic enum and value. A local function `Capture(topic, record)` closing over `session`/`streamId`/`working`/`now`/`topicsCaptured` would collapse the tail while leaving each typed property assignment (and PrimaryGoal's extra TargetEvent-clearing branch) inline. Subjective — the explicit per-topic form is also defensible.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/SubmitStructuredAnswersHandler.cs
```diff
@@ -0,0 +1,242 @@
...
+/// plan-generation terminal branch and append <see cref="PlanLinkedToUser"/> +
+/// <see cref="OnboardingCompleted"/>; (5) record the idempotency marker last. Any failure (e.g. a
+/// rejected generated plan) propagates so Wolverine aborts the transaction — nothing staged, the
+/// form is re-submittable.
+/// </para>
+/// </remarks>
+public sealed partial class SubmitStructuredAnswersHandler
+{
+    // The type is a stateless logical handler container: Wolverine codegen emits a non-static
+    // handler stub that calls the static Handle method, and ILogger<SubmitStructuredAnswersHandler>
```
leehopper: **💡 [LOW] Bare code-shaped tokens in prose comments (REVIEW.md backtick rule)**

Per REVIEW.md's code-comment rule, code-shaped tokens in prose should be backtick-wrapped to avoid tripping SonarAnalyzer S125. Line 42 uses a bare `ILogger<SubmitStructuredAnswersHandler>` and line 203 a bare `JsonDocument.Deserialize<T>()`. **Contested:** the build is green, so these specific mostly-prose comments likely don't actually fire S125 — this is style-rule adherence, not a build fix. Optional.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/SubmitStructuredAnswersRequestMapper.cs
```diff
@@ -0,0 +1,317 @@
...
+        {
+            normalized = null;
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
leehopper: Fixed in b1e1a91 — widened the catch in `TryNormalizeOptionalDuration` to `catch (Exception ex) when (ex is FormatException or OverflowException)`. `XmlConvert.ToTimeSpan` does checked arithmetic, so a syntactically valid but numerically huge `xsd:duration` (`P1000000000D`) threw `OverflowException`, which escaped `TryMap`'s `ArgumentException` catch and (because the controller calls `TryMap` outside its `try`) surfaced as a 500. Now returns a clean 400. Guarded by a mapper unit test (`TryMap_OverflowDuration_Fails`) and an end-to-end no-staging integration test (`SubmitAnswers_OverflowDuration_Returns400_WithoutStagingEvents`).

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Onboarding/OnboardingAnswersEndpointIntegrationTests.cs
```diff
@@ -0,0 +1,557 @@
...
+        profile!.PrimaryGoal.Should().Be(PrimaryGoal.RaceTraining);
+        profile.TargetEvent!.EventName.Should().Be("Berlin Marathon");
+        profile.WeeklySchedule!.MaxRunDaysPerWeek.Should().Be(5);
+        profile.CurrentPlanId.Should().Be(state.CurrentPlanId);
+        profile.OnboardingCompletedAt.Should().NotBeNull();
+
+        // Assert — a plan stream was actually created at the linked id.
+        (await PlanStreamExistsAsync(Factory, userId, state.CurrentPlanId!.Value, ct))
+            .Should().BeTrue(because: "the terminal branch starts a plan stream at the linked plan id");
+    }
```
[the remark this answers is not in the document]
leehopper: Fixed in b1e1a91 — `SubmitAnswers_AllSixTopics_…`, `SubmitAnswers_FitnessProfile_…`, and `SubmitAnswers_IncompleteSet_…` now assert the full `OnboardingStateDto` contract with a single `.Should().BeEquivalentTo(expected, …)`. The server-generated `CurrentPlanId` is excluded from the deep comparison and asserted separately (`NotBeNull().And.NotBe(Guid.Empty)` where a plan is expected; deterministic `null` otherwise). The `LoadProfileAsync`/`PlanStreamExistsAsync` side-effect checks are retained.

backend/tests/RunCoach.Api.Tests/Modules/Coaching/Onboarding/OnboardingAnswersEndpointIntegrationTests.cs
```diff
@@ -0,0 +1,557 @@
...
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+    }
+
+    [Fact]
+    public async Task SubmitAnswers_PresentTopicMissingRequiredField_Returns400()
+    {
+        // Arrange — a raw body with a weeklySchedule object missing its required fields. The loosened
+        // input DTO's [JsonRequired] presence enforcement must surface this as a clean 400 at model
+        // binding, never a 500.
```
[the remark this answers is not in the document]
leehopper: Fixed in b1e1a91 — wrapped the bare code-shaped tokens in backticks (`JsonRequired`; `ConversationView`, `OnboardingStarted`, `StartStream`, `ExistingStreamIdCollisionException`; `double`; `runcoach_events`). Release build is clean under `TreatWarningsAsErrors`.

backend/openapi/swagger.json
```diff
@@ -2687,6 +2894,43 @@
...
+          },
+          "injuryHistory": {
+            "$ref": "#/components/schemas/InjuryHistoryInputDto"
+          },
+          "preferences": {
+            "$ref": "#/components/schemas/PreferencesInputDto"
+          }
+        },
+        "additionalProperties": false
+      },
```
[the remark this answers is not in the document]
leehopper: Deferred — this is a known repo-wide artifact, not specific to this endpoint. `RequireNonNullablePropertiesSchemaFilter` marks every `$ref` property `required` regardless of C# nullability (40+ properties across the API, e.g. the shipping `MesoDaySlotOutput.workoutType`). Runtime binding correctly treats the six topic slots as optional — verified by the partial-submission integration tests (`SubmitAnswers_IncompleteSet_…`, `SubmitAnswers_FitnessProfile_NoTargetEvent_…`). The structural fix to the schema filter is tracked separately (PR #163); PR-D hand-writes accurate optional-topic client models (like `plan.model.ts`) rather than consuming the over-`required` generated schema. Patching only this one DTO would diverge from the repo-wide pattern.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/Models/SubmitStructuredAnswersRequestDto.cs
```diff
@@ -0,0 +1,28 @@
...
+/// <param name="InjuryHistory">InjuryHistory answer, or null.</param>
+/// <param name="Preferences">Preferences answer, or null.</param>
+public sealed record SubmitStructuredAnswersRequestDto(
+    [property: JsonRequired] Guid IdempotencyKey,
+    PrimaryGoalInputDto? PrimaryGoal,
+    TargetEventInputDto? TargetEvent,
+    CurrentFitnessInputDto? CurrentFitness,
+    WeeklyScheduleInputDto? WeeklySchedule,
+    InjuryHistoryInputDto? InjuryHistory,
+    PreferencesInputDto? Preferences);
```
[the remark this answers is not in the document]
leehopper: Not changing this one. A missing `IdempotencyKey` returns the framework-standard `ValidationProblemDetails` 400 — the app-wide model-binding contract every `[ApiController]` route produces — so it's still a 400. The semantic `invalid-onboarding-answers` problem type is reserved for validation the mapper performs, and the *present-but-empty* `Guid.Empty` case already routes through it (`TryMap` rejects `Guid.Empty`). Unifying the model-binding 400 would require an app-wide `InvalidModelStateResponseFactory` override, which would change the error contract for every other endpoint — out of scope for this additive endpoint.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/SubmitStructuredAnswersRequestMapper.cs
leehopper: Fixed in b1e1a91 — catch widened to `catch (Exception ex) when (ex is FormatException or OverflowException)`, exactly as suggested. Added `TryMap_OverflowDuration_Fails` (unit) and `SubmitAnswers_OverflowDuration_Returns400_WithoutStagingEvents` (integration) asserting `P1000000000D` → 400 with no events staged. Thanks for the re-anchor — this was the real one.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/OnboardingController.cs
leehopper: Fixed in b1e1a91 — the class `<summary>` now enumerates the fourth endpoint ("…revise an answer, and submit structured form answers").

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/OnboardingController.cs
leehopper: Acknowledged — not adding coverage. This is the provably-unreachable defensive branch you identified: the handler always bootstraps-then-appends, so the post-commit reload cannot be null. No controller-level unit harness exists, and forcing the branch would require mocking Marten to lie about its own just-committed write. Leaving the guard + Error-level log in place as defence-in-depth.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/SubmitStructuredAnswersHandler.cs
leehopper: Keeping the explicit per-topic form (you flagged it subjective). A `Capture(topic, record)` local function collapses the append/clear/increment tail, but each block's typed `working.X = cmd.X` assignment differs and `PrimaryGoal` carries the extra `RaceTraining` target-clearing branch — so it would need a per-topic assignment lambda, trading six flat blocks for six closures without a real readability gain.

backend/src/RunCoach.Api/Modules/Coaching/Onboarding/SubmitStructuredAnswersHandler.cs
leehopper: Fixed in b1e1a91 — backtick-wrapped `ILogger<SubmitStructuredAnswersHandler>` (line 42) and `JsonDocument.Deserialize<T>()` for consistency with the S125 comment rule, even though the build was already green.

PR #285, opened 2026-07-14
feat(split-ui): Slice 2 — Today (hero, THE WEEK, coach digest, THE BLOCK)

leehopper:
Recomposes home as **Today** — the daily-driver screen. Six sections in the design's locked order: header → workout hero → THE WEEK → FROM YOUR COACH → UP NEXT → THE BLOCK. Deletes the old `TodayCard` / `MacroPhaseStrip` / `MesoWeekBlock` / `UpcomingList` composition.

Slice 2 of the SPLIT / Alpine cycle. Depends on Slice 1's shell. **Not stacked** — branches from `main`.

## Commits

| | |
|---|---|
| **PR-A** | target-event fields on the plan projection + codegen |
| **PR-B** | header, workout hero, THE WEEK + the shared date/week derivations |
| **PR-C** | coach digest, adaptation-diff helper, composer receiver, UP NEXT |
| **PR-D** | THE BLOCK, furniture deletion, e2e realignment |
| **review** | fixes from a whole-diff review (below) |

## The one backend delta

`PlanGenerated` / `PlanProjectionDto` gain three **nullable** target-event fields (name, distance, date), sourced from the onboarding `TargetEvent` at the generation site. THE BLOCK's goal chip needs a structured source; `MacroPlanOutput` carries only free-prose `GoalDescription`, and DEC-089 D4 rejected `GET /onboarding/state` because it reflects last-onboarding answers rather than the active plan, and drifts on regenerate.

Marten streams are append-only, so events already in the store lack these properties. Verified empirically that old-shape `PlanGenerated` JSON deserializes with all three as null, no exception. General-fitness plans carry nulls and simply render no goal chip.

## Two things the design didn't cover, decided during build

**The hero didn't know about logs.** Logging today's run flipped THE WEEK's cell to a green "done" check while the hero above it still showed **LOG RUN** for the same workout — one run reading as done and not-done on one screen. The hero was built before the log data was wired in. Both surfaces now derive from a single `isDateLogged` predicate fed by one logs array and one `todayUtc` epoch, so they can't disagree *by construction* rather than by luck. The logged affordance reuses the same `--positive` token THE WEEK's done cell uses.

**`--alp-faint` was carrying essential text.** It's documented as decorative-only, fails AA by design, and is exempt from `check-contrast` — so nothing automated would ever have caught it. It had crept onto the hero's unit labels (`KILOMETERS` — essential under a km/miles preference: "10.0" alone is ambiguous), THE WEEK's rest-day labels, and THE BLOCK's phase-span labels (`BASE 1–4`, the runner's only way to know which weeks belong to which phase). All now use the AA-gated `--muted-foreground`. A deliberate, documented deviation from the mock. Its only surviving use is the `aria-hidden` chevron.

## Bugs caught before they shipped

- **A wrong date in the hero, in every non-UTC timezone.** The eyebrow mixed a local `getDay()` with a UTC `formatShortDateUtc` on the same `Date`. There is now exactly one notion of "today's calendar date" flowing through the screen.
- **The digest could silently grow.** Its `PLAN ADJUSTED` summary was unclamped; a realistic two-sentence headline wraps at 375px. The test meant to prevent this compared `getBoundingClientRect` heights under jsdom — which does no layout — so it was asserting `0 === 0` and could never fail.
- **`parseIsoDateUtc('2026-02-30')` returned March 2** instead of null, contradicting its own doc comment.
- **`regenerate-plan.spec.ts` asserted on a testid this PR deletes.** Realigned onto THE BLOCK's phase labels — same invariant. Both e2e plan fixtures were also missing `planStartDate`, without which THE WEEK's "done" and "today" cells were structurally unreachable in e2e.
- **A test titled "renders summary" never asserted the summary.**

Client week/phase/date math replicates `WeekContext`/`PlanCalendar` exactly, including a zero-week phase contributing nothing and not shifting later phases.

## Known, deliberate

- `isDateLogged` is a pure date match with no slot correlation — a log recorded today against an unrelated workout marks today done. Pre-existing behaviour of THE WEEK; the hero now shares it, so the two agree. Worth a product call later.
- A bonus run logged on a **rest day** shows done in THE WEEK but the rest hero doesn't acknowledge it. An asymmetry, not a contradiction (the rest hero claims nothing about logging).
- `MicroWorkoutCard` is now unmounted — its docstring says it's kept deliberately for a future detail-card screen. Left in place.

## Context: THE WEEK is about to become honest

The header's `WEEK N OF M` reads through `resolveCurrentWeek`'s clamp. That clamp has been masking a real bug — plan generation only ever emits 4 meso weeks and **one** micro week, so only week 1 of any plan has real workouts. Fixing that is #284 (DEC-090, rolling horizon), which lands separately. This slice needs no change to benefit; it just gets real data to clamp against.

## Verification

Backend 1986/1986 (real Testcontainers, incl. the new `GetPlanCurrent_EchoesTargetEventFields`). Frontend build clean, **836/836** vitest, eslint zero errors and zero warnings, `check-contrast` 38/38 in both themes, no codegen drift.

Playwright was **not executed** — it needs a running stack. The e2e realignment was traced statically against the new DOM.

https://claude.ai/code/session_013NXAXGXeAFKFyNm5mVTdQS

## Summary by CodeRabbit

* **New Features**
  * Added onboarding “target event” support (name, distance, date) and surfaced it in the plan goal chip when available.
  * Updated the Today plan UI to new sections: workout hero, weekly grid, coach digest, up next, and THE BLOCK overview.
  * Coach chat can now prefill and auto-focus the composer from navigation state, and coach digest shows adaptation-aware headlines.
* **Bug Fixes**
  * Improved UTC date/week boundary handling for today/progress behavior and tightened km/miles formatting and unknown/logs-unavailable states.


frontend/e2e/plan-render.spec.ts
```diff
@@ -2,7 +2,8 @@ import { randomUUID } from 'node:crypto'
 import { expect, test, type Page, type Route } from '@playwright/test'
 
 // Plan-render e2e per spec § Unit 4 R04.1, R04.3, R04.9 + the canonical
-// `plan-view-on-home.feature` Gherkin (slice 1).
+// `plan-view-on-home.feature` Gherkin (slice 1), realigned for the SPLIT/
+// Alpine Today-screen recomposition (Slice 2 §7).
```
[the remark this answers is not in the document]
leehopper: Fixed. Both comments rewritten to describe the behaviour inline; no `.feature` or `CLAUDE.md` paths remain.

frontend/e2e/regenerate-plan.spec.ts
```diff
@@ -47,6 +50,23 @@ const planAId = '8a4b9b2a-1d3f-4f1c-9aab-5e2c1f0b1234'
...
 const userId = '00000000-0000-0000-0000-000000000001'
 
+// The Sunday (LOCAL calendar day, matching `toUtcMidnight`'s
+// getFullYear/getMonth/getDate approach in `plan-display.helpers.ts`) on or
+// before "now", formatted `YYYY-MM-DD`. Both plan A and plan B share this
+// anchor (their `mesoWeeks`/`microWorkoutsByWeek` shapes are otherwise
+// identical week-1-4 templates) so THE WEEK's calendar-date join resolves
+// for either fixture — an absent/unparseable `planStartDate` makes every
+// day cell degrade to `planned`/`rest` (`resolveDayCells`'s unparseable-date
+// path), same fixture gap `plan-render.spec.ts` had.
```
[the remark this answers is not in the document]
leehopper: Fixed. The comment now states the local-calendar-day/date-join constraint directly, without naming a source file.

frontend/src/app/modules/coaching/components/adaptation-digest.helpers.ts
```diff
@@ -0,0 +1,105 @@
...
+// `before-after-diff.helpers.ts` stays untouched this slice — it keeps
+// backing the pre-Slice-3 `AdaptationTurn`/`BeforeAfterDiff` transcript card
+// on `/coach`. This file is additive, not a replacement.
+//
+// `formatChangeLocusDate` is built here, in this slice, purely so Slice 3 can
+// import it unchanged for its own "WHAT CHANGED" expander rows — Slice 2
+// itself never calls it (the digest card shows a headline sentence, never a
+// date). "One implementation of the weekNumber/dayOfWeek → calendar-date
+// locus math, not two" (Slice 2 spec §3) — do not let a future slice
+// re-derive this inside `before-after-diff.helpers.ts` instead.
```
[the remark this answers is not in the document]
leehopper: Fixed. Module header rewritten: the forward-reference to a later slice is gone, and the sibling-file names with it. It now just states that this is the single implementation of the weekNumber/dayOfWeek → calendar-date locus math and must not be re-derived elsewhere — a constraint that stands on its own.

frontend/src/app/modules/coaching/components/coach-digest.component.tsx
```diff
@@ -0,0 +1,231 @@
...
+  | { kind: 'restructure'; diff: PlanAdaptationDiffDto }
+  | { kind: 'plain'; userLine: string | null; coachLine: string | null }
+
+/**
+ * The fixed one-step lookback (Slice 2 spec §1 PR-C "userLine rule"): reads
+ * `turns[turns.length - 2]` as its own array element, never a field nested on
+ * the latest turn. Index arithmetic, not `Array.prototype.at` (ES2020
+ * target — see `plan-display.helpers.ts`'s `phaseForWeek` for the same
+ * convention).
+ */
```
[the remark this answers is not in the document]
leehopper: Fixed — the embedded file name is gone. Note I declined the adjacent nitpick proposing a `DigestCard` sub-component split plus a raw CSS var: it's a structural refactor of a component that just landed, and you tagged it optional yourself. Happy to revisit.

frontend/src/app/modules/coaching/components/conversation.fixture.ts
```diff
@@ -47,3 +53,133 @@ export const buildDiff = (
...
+// real, multi-element `turns` array (each user/coach message its own array
+// entry, never a single record carrying both — the doc comment on
+// `conversation.model.ts`'s `ConversationTimelineTurnDto` is ground truth
+// here) so the `userLine` one-step-lookback tests can control
+// `turns[turns.length - 2]` independently of `turns[turns.length - 1]`.
+// `coach-chat.component.spec.tsx` keeps its own unexported inline builders
+// (a different, page-level test surface) — these are the digest's own,
+// centralised here rather than re-inlined a third time.
+// ─────────────────────────────────────────────────────────────────────────
+
```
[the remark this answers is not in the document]
leehopper: Fixed.

frontend/src/app/modules/common/test-utils/render-in-both-themes.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const dark = render(ui, { container: darkContainer, baseElement: darkContainer })
+
+  const lightContainer = document.body.appendChild(document.createElement('div'))
+  root.classList.add('light')
+  const light = render(ui, { container: lightContainer, baseElement: lightContainer })
+
+  root.className = originalClassName
+
+  return { dark, light }
+}
```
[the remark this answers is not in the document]
leehopper: Fixed. While in here I also closed a real bug this file had: it leaked its mounted DOM containers across tests. It now tracks and tears them down in `afterEach`.

frontend/src/app/modules/plan/components/plan-display.helpers.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+  })
+
+  it('agrees on the same date even under a process timezone far BEHIND UTC', () => {
+    // A local read would roll backward into Tuesday July 7 under this
+    // offset (UTC-11).
+    process.env.TZ = 'Pacific/Niue'
+    const todayUtc = Date.UTC(2026, 6, 8)
+    expect(formatHeroEyebrowDate(todayUtc)).toBe('Wednesday, JUL 8')
+  })
+})
```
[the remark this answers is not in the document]
leehopper: Fixed. Also corrected a genuine test-pollution bug alongside it: teardown was assigning `undefined` to `process.env.TZ`, which coerces to the string "undefined" rather than clearing it, polluting every later test in the file. It now deletes the key when there was no original value.

frontend/src/app/modules/plan/components/plan-display.helpers.ts
```diff
@@ -102,10 +106,10 @@ export const DAY_OF_WEEK_LABELS: readonly string[] = [
...
 
 /**
- * Guard predicate the `TodayCard` and `UpcomingList` components share to
- * locate the workout matching a particular `dayOfWeek` (0-6). Returns
- * `undefined` when no workout is scheduled for that day — typical for the
- * runner's rest day.
+ * Guard predicate `home.page.tsx` uses to locate the workout matching a
+ * particular `dayOfWeek` (0-6) when constructing `WorkoutHero`'s
+ * `WorkoutHeroContent` (§2.1). Returns `undefined` when no workout is
+ * scheduled for that day — typical for the runner's rest day.
```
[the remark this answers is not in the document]
leehopper: Fixed.

frontend/src/app/modules/plan/components/workout-hero.component.tsx
```diff
@@ -0,0 +1,285 @@
...
+// F1 fix: `logged` is a 4th, sibling kind — not a boolean layered onto
+// `run` — carrying the exact same `slot`/`workout` shape `run` does. Today's
+// workout is either not-yet-logged (`run`) or already-logged (`logged`);
+// there is no state where `logged` is true without a `workout` to show, so a
+// distinct kind (rather than an `isLogged` flag that could theoretically be
+// `true` while `workout` is somehow absent) keeps that impossible state
+// unrepresentable, the same discipline the original C3 fix established.
+// `home.page.tsx`'s `resolveHeroContent` picks `logged` vs. `run` from
+// `isTodayLogged`, itself derived from THE WEEK's own log-join predicate
+// (`isDateLogged`, `the-week.helpers.ts`) — see that function's doc comment.
```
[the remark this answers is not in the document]
leehopper: Fixed. This one was the worst offender — the comments carried internal review labels ("C3 fix", "F1 fix") from my own fix loop, which mean nothing to anyone reading the file later. All purged, and the load-bearing rationale (why the stat band needs its own variant, why the content union is discriminated) is now stated inline.

frontend/src/app/modules/plan/pages/home.page.tsx
```diff
@@ -1,36 +1,65 @@
...
+ *   - On 200 renders all six sections.
  *   - On 404 renders the "no plan yet" defensive state with a CTA back
  *     to `/onboarding` — should not appear in normal flow because the
  *     top-level route guard catches the unborn-plan case first
  *     (spec § Unit 4 R04.3).
  *   - On other errors renders a generic failure surface.
  *
- * Spec § Unit 4 R04.1, R04.3, R04.4, R04.5, R04.6, R04.7.
+ * Spec § Unit 4 R04.1, R04.3, R04.4, R04.5, R04.6, R04.7 (Slice 1); Slice 2
+ * §1 PR-D for the six-section recomposition.
```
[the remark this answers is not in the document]
leehopper: Fixed — the spec-section citations are gone. I kept the `sonarjs/no-nested-conditional` rule-name mention: naming the lint rule the code is deliberately structured around is a durable fact about the code, not a pointer into a planning doc. Say the word if you'd rather it went too.

leehopper:
@coderabbitai resolve

frontend/src/app/modules/plan/pages/home.page.tsx
```diff
@@ -72,37 +96,133 @@ interface PlanLayoutProps {
...
+
+  const workout: MicroWorkoutCardDto | undefined =
+    slot?.slotType === 'Run' ? findWorkoutForDay(currentWeekWorkouts, today.getDay()) : undefined
+
+  const nextWorkout = findNextWorkoutAfter(currentWeekWorkouts, today.getDay())
+
+  // SAME predicate, SAME `logs` array, SAME `todayUtc` epoch `TheWeek` below
+  // uses for its today cell's `done` state — never a second, parallel "is
+  // today logged" check re-derived here.
+  const isTodayLogged = isDateLogged(logs, todayUtc)
```
leehopper: **🟠 [HIGH] Home page silently treats a failed/in-flight workout-log fetch as "no logs logged"**

`PlanLayout` destructures only `data` from `useGetWorkoutLogHistoryInfiniteQuery(undefined)` and falls back to an empty array. The `isError`/`isLoading` flags are never consumed, so a failed history request (5xx, network drop, auth expiry) is indistinguishable from "the runner has logged nothing". The empty `logs` array then feeds `isDateLogged(logs, todayUtc)` (which returns false) and `TheWeek`, so the hero renders the "Log run" CTA for a day the runner has ALREADY logged, THE WEEK renders zero `done` cells, and the progress bar reads `0.0/NN.N KM` — all of them presenting fabricated data as fact rather than an unknown/error state, with no logging, toast, or error surface anywhere on the path. The concrete user consequence is a duplicate log: the hero invites the runner to log a run that is already recorded. The sibling consumer of the exact same hook, `history.page.tsx` (lines 100-125), threads both `isLoading` and `isError` into `HistoryBody` and renders an error surface — this new call site diverges from that established convention. The same silent-degradation applies during the in-flight window, where the hero briefly shows "Log run" before flipping to "Logged".

**Suggested fix:** Destructure `isError` (and `isLoading`) from the infinite query and thread them into `WorkoutHero`/`TheWeek` so the logged-state affordances degrade to an explicit unknown/error treatment instead of asserting "not logged". At minimum, suppress the `logged`-vs-`run` decision (and the `0.0` progress claim) when the log fetch has not succeeded, and surface the failure the way `HistoryPage` already does.

<sub>`bug-1` · bug-detector · validator 82% · challenge 85%</sub>

frontend/src/app/modules/coaching/components/coach-digest.component.tsx
```diff
@@ -0,0 +1,230 @@
...
+export const CoachDigest = ({ currentWeek, units, className }: CoachDigestProps): ReactElement => {
+  const { data } = useGetConversationTimelineQuery(undefined)
+  const body = resolveDigestBody(data?.turns ?? [])
+
+  if (body.kind === 'empty') {
+    return <EmptyState className={className} />
+  }
+
+  return <DigestCard body={body} currentWeek={currentWeek} units={units} className={className} />
```
leehopper, line 273: **🟡 [MEDIUM] CoachDigest renders the "Nothing yet" empty state when the timeline request fails**

`CoachDigest` destructures only `data` from `useGetConversationTimelineQuery(undefined)` and passes `data?.turns ?? []` into `resolveDigestBody`, which returns `{ kind: empty }` for an empty array. A failed timeline fetch (5xx, network error, expired session) therefore renders the empty state, which affirmatively tells the runner "Nothing yet. Tell me how training feels, or hand me a run to log." and offers cold-start prefill chips — i.e. the UI states as fact that the runner has no coaching history when the request simply did not succeed. The `restructure` path is worse in kind: a PLAN ADJUSTED headline that the runner never sees because the fetch failed, with no retry affordance and no logging on the path. The error is swallowed entirely — nothing distinguishes "empty conversation" from "could not load conversation".

**Suggested fix:** Destructure `isError` (and `isLoading`) and give the digest a distinct third branch — a quiet "Could not load your coach" line with a retry, or simply render nothing/a skeleton — instead of collapsing failure into the cold-start empty state.

<sub>`bug-2` · bug-detector · validator 88% · challenge 85%</sub>

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
```diff
@@ -196,7 +221,13 @@ export const CoachChat = (): ReactElement => {
...
         />
       )}
-      <CoachComposer onSend={send} isStreaming={isStreaming} />
+      <CoachComposer
+        key={location.key}
+        onSend={send}
+        isStreaming={isStreaming}
+        initialValue={locationState?.prefill ?? ''}
+        autoFocus={Boolean(locationState?.prefill) || Boolean(locationState?.focusComposer)}
+      />
```
leehopper, line 238: **🟡 [MEDIUM] key={location.key} on CoachComposer discards an in-progress draft when the user taps the already-active Coach tab**

CoachChat now remounts CoachComposer via key={location.key}. React Router regenerates location.key on every navigation, including a same-URL navigation: useLinkClickHandler converts a click on a Link/NavLink whose resolved path equals the current path into a REPLACE, and the history package mints a fresh key on replace just as it does on push. The unmodified TabBar (`tab-bar.component.tsx` line 22 — `NavLink to=/coach`, no `replace` prop) is therefore a live trigger: a runner who is already on /coach, has typed a message into the composer, and taps the COACH tab gets a new location.key, a remounted CoachComposer, and a fresh useState(initialValue) — the typed draft is silently discarded. Before this PR the composer had no key and its draft survived any re-render. The prefill/focus contract only needs a fresh instance when router state actually carries prefill/focusComposer, so keying on every location.key is broader than the requirement.

**Suggested fix:** Key the composer on something that only changes when a seed is actually delivered — e.g. key={locationState === null ? coach-composer : location.key} — so plain navigations (including tapping the active Coach tab) leave the mounted composer, and its draft, intact.

<sub>`cross-file-1` · cross-file-impact · validator 85% · challenge 88%</sub>

frontend/src/app/modules/coaching/components/adaptation-digest.helpers.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+export function formatChangeLocusDate(params: {
+  planStartDate: string
+  weekNumber: number
+  dayOfWeek: number
+}): string | null {
+  const { planStartDate, weekNumber, dayOfWeek } = params
+  const date = resolveCalendarDateUtc(planStartDate, weekNumber, dayOfWeek)
+  return date === null ? null : formatShortDateUtc(date)
+}
```
leehopper: **🟡 [MEDIUM] formatChangeLocusDate is a new exported helper with no production caller**

The new `formatChangeLocusDate` export has zero production consumers — a repo-wide grep finds it only in its own doc comment and in `adaptation-digest.helpers.spec.ts`. The file header states the fact outright (the digest card itself never calls it, it shows a headline sentence, never a date), and `CoachDigest` only calls `composeAdaptationHeadline`. The REVIEW.md AI-generated-code rule asks to check for helpers created for operations that never happen; this one is a two-line wrapper over `resolveCalendarDateUtc` + `formatShortDateUtc` that exists only to be tested. Deleting it removes ~19 lines of source plus its spec block with no behavioural effect; a later slice that needs a date fragment can compose the two shared primitives at the point of use.

**Suggested fix:** Delete `formatChangeLocusDate` (lines 80-98), the sentence about it in the file header comment, and its describe block in the spec. Before: `export function formatChangeLocusDate(params) { const date = resolveCalendarDateUtc(...); return date === null ? null : formatShortDateUtc(date) }` with no caller. After: no such export; a future caller composes `resolveCalendarDateUtc` + `formatShortDateUtc` directly.

_Improvement suggestion — not a blocking defect._

<sub>`simplify-1` · code-simplifier · validator 70% · challenge 93%</sub>

frontend/src/app/modules/plan/components/the-week.helpers.ts
```diff
@@ -0,0 +1,142 @@
...
+  const weekStartEpoch = weekStart.getTime()
+  const weekEndEpoch = weekEnd.getTime()
+
+  const matching = logs.filter((log) => {
+    const epoch = parseIsoDateUtc(log.occurredOn)
+    return epoch !== null && epoch >= weekStartEpoch && epoch <= weekEndEpoch
+  })
+  const totalMeters = matching.reduce((sum, log) => sum + log.distanceMeters, 0)
+  return totalMeters / METERS_PER_KM
+}
```
leehopper, line 130: **🟡 [MEDIUM] weekLoggedKm inclusive Sunday/Saturday span boundaries are never tested**

weekLoggedKm filters logs with an inclusive epoch comparison (epoch >= weekStartEpoch && epoch <= weekEndEpoch) against the week Sunday and Saturday anchors. The spec only feeds logs dated 2026-04-20 (Monday) and 2026-04-24 (Friday) — both strictly inside the span — plus one out-of-span log on the NEXT week Sunday. No test places a log on the span endpoints themselves (Sunday 2026-04-19 or Saturday 2026-04-25), so both inclusive comparisons could be silently weakened to strict inequalities and the whole suite would still pass. This directly backs the THE WEEK progress bar the runner reads to decide whether they hit their weekly volume.

**Suggested fix:** Add a boundary test: weekLoggedKm({ weekNumber: 1, planStartDate: PLAN_START_DATE, logs: [log(2026-04-19, 4000), log(2026-04-25, 4000), log(2026-04-18, 9000), log(2026-04-26, 9000)] }) and assert the total is exactly 8 km — proving both endpoints are included and both adjacent days excluded.

_Improvement suggestion — not a blocking defect._

<sub>`test-3` · test-analyzer · validator 70% · challenge 93%</sub>

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
```diff
@@ -37,6 +37,29 @@ import { TranscriptScroller } from './transcript-scroller.component'
...
+// any shape a caller passes to `navigate(to, { state })`), so a plain `as`
+// cast here would silently accept a differently-shaped `state` from a future
+// caller. This guard actually checks the two optional fields' types before
+// the cast-free narrowing below trusts them.
+const isCoachChatLocationState = (value: unknown): value is CoachChatLocationState =>
+  typeof value === 'object' &&
+  value !== null &&
+  (!('prefill' in value) || typeof (value as { prefill?: unknown }).prefill === 'string') &&
+  (!('focusComposer' in value) ||
+    typeof (value as { focusComposer?: unknown }).focusComposer === 'boolean')
```
leehopper, line 61: **🟡 [MEDIUM] isCoachChatLocationState rejection path (malformed router state) has no test**

The new isCoachChatLocationState type guard exists specifically to reject a differently-shaped router state before it is trusted — its own comment says a plain as-cast would silently accept a future caller passing the wrong shape. The CoachChat spec only exercises the three well-formed cases (prefill string, focusComposer true, null state). No test passes a malformed state (for example { prefill: 42 } or a bare string), so the guard rejection branch is entirely uncovered: the guard could be deleted and replaced by a cast and every test would still pass, which defeats the purpose of writing it.

**Suggested fix:** Add a test: locationMock.mockReturnValue({ state: { prefill: 42 }, key: nav-1 }); render CoachChat; assert the textarea has value empty-string and does not have focus. Optionally add a second case with a non-object state (e.g. state: some-string).

_Improvement suggestion — not a blocking defect._

<sub>`test-4` · test-analyzer · validator 70% · challenge 93%</sub>

frontend/src/app/modules/plan/pages/home.page.tsx
```diff
@@ -72,37 +96,133 @@ interface PlanLayoutProps {
...
+    return { kind: 'unavailable' }
+  }
+  if (slot.slotType !== 'Run') {
+    return { kind: 'rest', slot, nextWorkout }
+  }
+  if (workout === undefined) {
+    return { kind: 'unavailable' }
+  }
+  return isTodayLogged ? { kind: 'logged', slot, workout } : { kind: 'run', slot, workout }
+}
```
leehopper, line 133: **🟡 [MEDIUM] resolveHeroContent rest-day and unavailable branches are never exercised at the page level**

resolveHeroContent is the page-local function that maps (slot, workout, nextWorkout, isTodayLogged) onto the hero discriminated union. Every HomePage test pins the system clock to Monday 2026-04-20, which the week-1 fixture template marks as a Run slot, so only the run and logged branches are exercised. The rest branch (slot.slotType !== Run, the runner rest day, several days a week in production) and both unavailable branches (currentWeekTemplate undefined, and a Run slot whose workout is missing — which per the known DEC-090 issue is the ACTUAL production state for every week past week 1) have no page-level coverage. WorkoutHero has its own spec, but it is handed a pre-built union member; nothing tests that the page picks the right member, including that the rest branch threads nextWorkout (from findNextWorkoutAfter) rather than the wrong workout.

**Suggested fix:** Add two HomePage tests: (1) setSystemTime to Sunday 2026-04-19 (Rest slot in the week-1 fixture) and assert workout-hero dataset.variant is rest and that workout-hero-next-workout names the Monday easy run; (2) render a plan whose current week template has a Run slot but whose micro workouts array is empty, and assert the hero renders the unavailable variant rather than crashing or claiming a workout.

_Improvement suggestion — not a blocking defect._

<sub>`test-2` · test-analyzer · validator 63% · challenge 88%</sub>

backend/src/RunCoach.Api/Modules/Training/Plan/PlanGenerated.cs
```diff
@@ -43,4 +57,7 @@ public sealed record PlanGenerated(
     DateOnly PlanStartDate,
     string PromptVersion,
     string ModelId,
-    Guid? PreviousPlanId);
+    Guid? PreviousPlanId,
+    string? TargetEventName,
+    double? TargetEventDistanceKm,
+    DateOnly? TargetEventDate);
```
leehopper, line 63: **🟡 [MEDIUM] Event-model change contradicts the cycle plan constraint, and the spec-mandated addendum was never written**

The SPLIT/Alpine cycle plan states the backend deltas for the whole cycle are additive and that there are no event-model changes beyond the additive AnswerCaptured field. This PR appends three positional parameters to `PlanGenerated`, a persisted Marten event record — an event-model change outside that carve-out. The slice spec itself anticipated the conflict and recorded it as open question #1, concluding it is can-pick-default and not blocking, but explicitly saying it is worth a one-line addendum to the cycle-plan or a decision-log note recording the interpretation. The PR contains zero documentation changes (no .md files in the 62-file diff), so neither the cycle-plan addendum nor the decision-log note was written. The chosen approach is defensible; the missing record is the gap — a future reader of the cycle plan will still believe the event model was untouched this cycle.

**Suggested fix:** Add the one-line cycle-plan addendum (or a decision-log entry) the spec calls for: record that Slice 2 appends three nullable fields to `PlanGenerated`, why `MacroPlanOutput` was rejected as the carrier, and that previously-stored events hydrate with nulls so no upcaster is required.

<sub>`conv-1` · conventions-and-intent · validator 40% · challenge 86%</sub>

frontend/src/index.css
```diff
@@ -259,6 +273,22 @@
...
+   * 1.4.11 decorative/non-text contrast exemption as `--border` above.
+   * Text usages of clay use `--clay-text` above instead (AA-verified for
+   * text; `--clay-marker`'s light value is a different,
+   * untested-for-contrast color). */
+  --clay-marker: var(--alp-clay-marker);
+  /* THE BLOCK's `distant`-tier cell fill — a genuinely new primitive (no
+   * existing token matched the mock's dimmer fill), border/fill-only
+   * (never text — the phase-span label row sits below the cell grid, never
+   * layered over a cell), so it carries the same contrast exemption as
+   * `--border`/`--clay-marker`. */
```
leehopper, line 290: **💡 [LOW] Two new never-contrast-checked tokens added without updating the exempt list documented in frontend/CLAUDE.md**

frontend/CLAUDE.md enumerates the check-contrast exempt set as exactly three tokens. This PR adds two new semantic tokens, `--clay-marker` and `--surface-dim`, neither of which is added to check-contrast assertions — the `check-contrast.ts` change is comment-only, extending its in-file EXEMPT list to five entries. The CLAUDE.md styling section, which is the convention file a future contributor reads, still names only three, so it now understates the un-gated surface: a reader would reasonably assume any other semantic token is machine-checked. Since the exemption for both new tokens rests entirely on the claim that they are border/fill-only and never carry text, that claim needs to live where the rule lives.

**Suggested fix:** Update frontend/CLAUDE.md § Design tokens & theming to list `--clay-marker` and `--surface-dim` alongside the other exempt tokens, with the same border/fill-only justification, so the convention doc and the script agree on the un-gated set.

_Improvement suggestion — not a blocking defect._

<sub>`conv-6` · conventions-and-intent · validator 65% · challenge 88%</sub>

frontend/src/app/modules/common/components/stat-band/stat-band.component.tsx
```diff
@@ -30,14 +59,40 @@ export interface StatBandProps {
...
+   * `1fr/1.7fr/1fr` CSS grid with a full top+bottom hairline border.
+   */
+  variant?: 'divided' | 'hero'
 }
 
 /**
  * Horizontal row of {@link StatCell}s separated by 1px hairline dividers.
  * Used for the hero stat band (distance / pace / reps or duration).
+ *
+ * `divide-x divide-border` is reused unchanged for both variants — Tailwind's
```
leehopper, line 74: **💡 [LOW] StatBand doc comment describes the Tailwind v4 divide-x mechanism incorrectly**

The new `StatBand` doc comment justifies reusing `divide-x` across the flex and grid variants by claiming Tailwind applies a border-left to every child but the first via a `:not(:first-child)` sibling selector. That is not what the installed Tailwind v4 emits: the compiled rule is `:where(& > :not(:last-child))` setting `border-inline-end-width` — a border on the INLINE-END side of every child except the LAST. The conclusion (that the divider contract survives the flex-to-grid swap) still holds, but the stated mechanism is wrong, and a future reader relying on it to reason about first-child/last-child edge cases or RTL behaviour (border-inline-end, not border-left) would be misled by a comment that exists specifically to be the authority on this.

**Suggested fix:** Reword to the actual mechanism: Tailwind v4 `divide-x` targets `:where(& > :not(:last-child))` and sets `border-inline-end-width`, so each child except the last gets a trailing hairline — a child-scoped rule that is agnostic to whether the parent is flex or grid, which is why the variant swap is safe.

_Improvement suggestion — not a blocking defect._

<sub>`conv-3` · conventions-and-intent · validator 63% · challenge 93%</sub>

frontend/src/app/modules/plan/components/plan-display.helpers.ts
```diff
@@ -185,3 +209,243 @@ export const dayOfWeekIndex = (date: Date): number => date.getDay()
...
+ * established null-value convention.
+ */
+export function resolveHeroThirdStat(workout: MicroWorkoutCardDto): HeroThirdStat {
+  const workSegment = workout.segments.find((segment) => segment.repetitions > 1)
+  if (workSegment !== undefined) {
+    return {
+      value: `×${workSegment.repetitions}`,
+      label: `Reps · ${workSegment.durationMinutes} min`,
+    }
+  }
```
leehopper, line 440: **💡 [LOW] resolveHeroThirdStat doc claims an interval/repetition-style segment but the code matches ANY segment with repetitions > 1**

The doc comment says the third hero stat renders a rep count when the workout has an interval/repetition-style segment. The code does not test segment type at all — it takes the first segment with `repetitions > 1` and names the local `workSegment`. The same file already refuses to make that assumption: `describeSegment` guards `Recovery` for `repetitions > 1` explicitly, with a comment noting the wire contract statement about reps is made about `Work` specifically and is not proven to exclude `Recovery`. So the two comments in one file disagree: one says a Recovery segment may legitimately carry reps, the other implies only interval/repetition segments can. If a workout ever carries a repeated Recovery segment before a repeated Work segment, the hero would label the recovery reps as the workout reps.

**Suggested fix:** Either filter on segment type (`segment.segmentType === Work && segment.repetitions > 1`) so the code matches the comment, or reword the comment to say it takes the first segment carrying repetitions, whatever its type, and rename `workSegment` accordingly.

_Improvement suggestion — not a blocking defect._

> ⚠️ **Contested.** The blind challenger scored this claim only 52% likely correct: resolveHeroThirdStat does `workout.segments.find((s) => s.repetitions > 1)` with no segmentType test, names the result workSegment, and its doc says the branch fires 'when the workout has an interval/repetition-style segment' - while describeSegment in the same file explicitly handles Recovery with repetitions > 1 and has a passing spec for it. WorkoutSegmentOutput.Repetitions is a plain required int on every segment regardless of type, and no backend validator constrains reps>1 to Work. So the file holds two contradicting assumptions and the wire permits the bad case. But no code path produce

<sub>`conv-5` · conventions-and-intent · validator 59% · challenge 52%</sub>

frontend/src/app/modules/plan/components/the-week.helpers.ts
```diff
@@ -0,0 +1,142 @@
...
+ * reading of "a log exists for that date," matching {@link resolveDayCells}'s
+ * own rule. Sharing this one function (never a second, parallel "is today
+ * logged" check re-derived at the hero's mount site) is what makes it
+ * IMPOSSIBLE for THE WEEK's today cell and the hero's logged-state affordance
+ * to disagree about the same date — the exact bug this predicate's
+ * extraction closes.
+ */
+export function isDateLogged(logs: readonly WorkoutLogDto[], dateEpoch: number): boolean {
+  return logs.some((log) => parseIsoDateUtc(log.occurredOn) === dateEpoch)
+}
```
leehopper, line 50: **💡 [LOW] isDateLogged doc asserts an IMPOSSIBLE-to-disagree invariant the code does not guarantee**

The `isDateLogged` doc comment claims that sharing this one predicate makes it IMPOSSIBLE for THE WEEK day cell and the hero logged-state affordance to disagree about the same date. Sharing the predicate only guarantees the same answer for the same input date — but the two call sites feed it different dates. The hero calls `isDateLogged(logs, todayUtc)` with the real wall-clock day; `resolveDayCells` calls it with cell epochs derived from `resolveCalendarDateUtc(planStartDate, currentWeekNumber, day)`. Those diverge in at least two reachable cases: (a) `planStartDate` is unparseable, so every cell date is null and no cell can be done or today while the hero still renders logged; (b) `resolveCurrentWeek` clamps to the last populated week (the known 4-meso-week constraint means every runner past week 4 hits this), so the rendered week span does not contain today, no cell is today, and a run logged today shows nowhere in the grid while the hero shows Logged. The same absolute claim is repeated in `workout-hero.component.tsx` and `home.page.tsx`. Comments that assert an invariant this strongly get trusted by the next editor; this one overstates what the shared predicate buys.

**Suggested fix:** Soften the comment to what is actually guaranteed — one predicate, so the two surfaces cannot disagree WHEN THEY ARE ASKING ABOUT THE SAME DATE — and state explicitly that when `planStartDate` is unparseable or the current week is clamped away from today, the grid simply has no cell for today while the hero still reflects the real day.

_Improvement suggestion — not a blocking defect._

> ⚠️ **Contested.** The blind challenger scored this claim only 35% likely correct: The behavioural divergence is real and constructible: home.page.tsx:180 calls isDateLogged with the wall-clock todayUtc while resolveDayCells calls it with resolveCalendarDateUtc(planStartDate, currentWeek, day), and resolveCurrentWeek clamps to the last generated meso week, so past the plan's horizon a today-log shows in the hero but no grid cell is done or today. BUT the comment's actual wording carries the qualifier that saves it: it says impossible to disagree 'about the same date' - which a shared pure predicate does guarantee. The divergence cases are ones where the two surfaces evaluate

<sub>`conv-4` · conventions-and-intent · validator 57% · challenge 35%</sub>

leehopper (commented):
## Deep Review — 14 findings

Multi-agent review of PR #285 (SHA `933e6045`). 7 concern-parallel discovery agents (Opus / frontier tier) → deterministic verification → independent Sonnet validation → blind Opus challenge on every surviving finding.

**25 findings discovered → 11 filtered → 14 delivered.** None were removed by the blind challenge.

| Severity | Count |
|---|---|
| 🟠 High | 1 |
| 🟡 Medium | 7 |
| 💡 Low | 6 |

**Security: clean.** No new endpoints, no HTML sinks, `targetEventName` reaches a React text node (auto-escaped), the e2e fixture password predates this diff.

The one finding worth blocking on is **F-01** (`home.page.tsx`): the home page destructures only `data` from the workout-log query, never `isError`/`isLoading`. A failed log fetch is indistinguishable from "no logs", so the hero renders **LOG RUN for a day already logged** and THE WEEK shows `0.0` progress. This is the same class of contradiction the PR body describes fixing — hero and week disagreeing about one run — reintroduced via the error path rather than the date path. The sibling `history.page.tsx` already threads both flags into an error surface.

Two findings are **contested** (challenge score < 75) and flagged inline — treat them as judgment calls, not defects.

<details>
<summary><b>Review methodology</b></summary>

- **Agents (Phase 3, Opus):** bug-detector, security-reviewer, cross-file-impact, test-analyzer, conventions-and-intent, type-design-analyzer, code-simplifier
- **Validation (Phase 5, Sonnet):** 7 batch validators, all 25 findings independently assessed
- **Blind challenge (Phase 7, Opus):** 14 fresh agents, one per finding, given only the claim + raw code — never the original reasoning
- **Pipeline correction:** `verify_findings.py`'s git-blame lookup failed silently (returned `author: null`) and defaulted **11 of 14 findings to "surfaced" / pre-existing**. Re-deriving origin from the diff showed all 11 sit on lines this PR *added*. That error had downgraded their severity and instructed challengers to "rate confidence low — this code pre-dates the PR". Origin was corrected, severity restored, and **all 11 were re-challenged with honest framing**; scores rose materially (e.g. 55→88, 42→85, 62→93).
- **Not verified:** Playwright was not executed (per the PR body); the e2e realignment remains statically traced only.

</details>

🤖 Generated by deep-review · Reviewed up to: `933e60454c30d15af61207d1d65a29eab1a42470`

frontend/src/app/modules/plan/components/plan-display.helpers.ts
```diff
@@ -8,6 +8,8 @@
 // these maps to assert trademark cleanliness; behaviour belongs in the
 // `.component.tsx` files.
```
leehopper: **🟡 [MEDIUM] File-header invariant "Keep this file string-literal-only" is contradicted by the 282 lines of logic this PR adds to it**

The header comment at the top of `plan-display.helpers.ts` states a rule for the file: keep it string-literal-only, with behaviour living in the `.component.tsx` files. This PR adds ~282 lines of behaviour to exactly this file — the whole UTC date pipeline (`toUtcMidnight`, `parseIsoDateUtc`, `resolveCalendarDateUtc`, `formatShortDateUtc`, `formatHeroEyebrowDate`), the summary composer (`describeSegment`, `composeWorkoutSummary`), and the hero stat derivations (`resolveHeroThirdStat`, `resolveHeroDistanceStat`) — turning it into the canonical logic module for the whole Today screen. The design change is reasonable, but the file still carries an invariant comment that now flatly contradicts its own contents, and the comment was left untouched in a PR whose final commit was specifically about comment discipline. Either update the header to describe what the module now is, or move the date/summary logic into its own helpers module.

**Suggested fix:** Rewrite the header to describe the module as it now stands (shared display constants plus the canonical UTC date pipeline and hero summary/stat derivations for the plan module), or split the date pipeline into a separate `plan-date.helpers.ts` and keep the string-literal-only invariant true.

_Improvement suggestion — not a blocking defect._

<sub>`conv-2` · conventions-and-intent · validator 67% · challenge 85% · re-anchored to line 8 (the header comment's original line 1 is outside the diff hunks)</sub>

leehopper (commented):
Two findings from the deep review could not be anchored inline because their target lines fall outside this PR's diff hunks. Posting the one that can be re-anchored here; the other (`PlanProjectionDto.cs:86-97`) follows as a top-level comment.

leehopper:
**💡 [low] PlanProjectionDto XML docs still name MacroPhaseStrip and MesoWeekBlock, both deleted by this PR**

_Posted as a top-level comment: the target lines (`PlanProjectionDto.cs:86-97`) are outside this PR's diff hunks, so GitHub rejects an inline anchor there. The file itself is in the diff._

This PR deletes macro-phase-strip.component.tsx and meso-week-block.component.tsx, but the backend contract DTO it also edits still documents its two largest properties in terms of those now-nonexistent components: Macro is documented as the macro phase strip on the home page, and MesoWeeks is documented as rendered directly by MesoWeekBlock. The author did update the equivalent stale reference in units-toggle.component.tsx (TodayCard/UpcomingList/MesoWeekBlock became WorkoutHero/TheWeek/UpNext/TheBlock), so this is a missed site of the same deletion sweep, and it leaves the shared API contract documentation pointing at components a reader cannot find. REVIEW.md requires productionized, self-contained comments.

**Suggested fix:** Update the two XML doc summaries to name the surviving consumers (TheBlock for Macro, TheWeek/TheBlock for MesoWeeks), matching the fix already applied in units-toggle.component.tsx.

A blind challenger also noted the drift is slightly broader than reported: `PlanRenderingController.cs:20` carries the same stale "macro phase strip" phrasing.

<sub>`cross-file-2` · cross-file-impact · validator 85% · challenge 85%</sub>

frontend/src/app/modules/plan/pages/home.page.tsx
leehopper: Fixed in cfa2f1a. `PlanLayout` now destructures `isLoading`/`isError` from the history query and derives `logsUnavailable = isLogsLoading || isLogsError`. `isTodayLogged` becomes `!logsUnavailable && isDateLogged(logs, todayUtc)`, and `logsUnavailable` is threaded into both `WorkoutHero` and `TheWeek`: the hero never renders the `logged` variant on an untrusted fetch and carries a quiet disclaimer line instead of letting LOG RUN imply "confirmed not logged"; THE WEEK forces an empty log list into both derivations (so no fabricated `done` cells) and renders `—/NN.N KM` rather than a fabricated `0.0`. Two new page tests cover the error and in-flight paths.

frontend/src/app/modules/coaching/components/coach-digest.component.tsx
leehopper: Fixed in cfa2f1a. `CoachDigest` now destructures `isLoading`/`isError` and returns distinct branches before `resolveDigestBody` is ever called: a quiet neutral placeholder while in flight, and a muted "Couldn't load your coach." line on failure. Neither reaches the cold-start copy or the prefill chips. Both branches have spec coverage asserting the cold-start text and chips are absent.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
leehopper: Fixed in cfa2f1a. The composer key is now `locationState === null ? 'coach-composer' : location.key` — stable across plain navigations (including the same-URL REPLACE from tapping the active COACH tab), fresh only when router state actually carries a prefill/focus seed. A new test types a draft, re-renders with `state: null` and a new `location.key`, and asserts the draft survives. Also corrected a now-stale claim in the composer's own docstring that said the parent remounts it on every navigation.

frontend/src/app/modules/coaching/components/adaptation-digest.helpers.ts
leehopper: Fixed in cfa2f1a. Confirmed zero production callers by repo-wide grep, then deleted the export, its JSDoc, the file-header sentence describing it, the now-unused imports, and its spec describe block.

frontend/src/app/modules/plan/components/the-week.helpers.ts
leehopper: Fixed in cfa2f1a. Added the boundary test: logs on 2026-04-19 (Sunday span start) and 2026-04-25 (Saturday span end) at 4000 m each, plus 2026-04-18 and 2026-04-26 at 9000 m each just outside, asserting the total is exactly 8 km. Weakening either `>=`/`<=` to a strict inequality now fails.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
leehopper: Fixed in cfa2f1a. Added two tests exercising the guard's rejection path — a wrong-typed field (`{ prefill: 42 }`) and a non-object state (a bare string) — each asserting the composer renders empty and unfocused. A plain cast would now fail them.

frontend/src/app/modules/plan/pages/home.page.tsx
leehopper: Fixed in cfa2f1a. Added two page-level tests: the clock pinned to Sunday 2026-04-19 (a Rest slot) asserts the hero variant is `rest` and that the next-workout row names Monday's easy run — proving the rest branch threads `nextWorkout` from `findNextWorkoutAfter`; and a Run-slot day whose micro-workouts array is empty asserts the `unavailable` variant rather than a crash or a claimed workout.

backend/src/RunCoach.Api/Modules/Training/Plan/PlanGenerated.cs
leehopper: Fixed in cfa2f1a. Added the addendum to the cycle plan's § Captured During Cycle table: Slice 2 appends three nullable target-event fields to the persisted `PlanGenerated` event; `MacroPlanOutput` was rejected as the carrier because it is an LLM structured-output schema carrying only free-prose `GoalDescription`; `GET /onboarding/state` was rejected per DEC-089 D4 (reflects last-onboarding answers, not the active plan, and drifts on regenerate); and previously-stored events hydrate with nulls on replay, so no upcaster is required.

frontend/src/index.css
leehopper: Fixed in cfa2f1a. `frontend/CLAUDE.md` § Design tokens & theming now lists all five exempt tokens, adding `--clay-marker` and `--surface-dim` with the border/fill-only-never-carries-text justification, so the convention doc and the script agree on the un-gated set.

frontend/src/app/modules/common/components/stat-band/stat-band.component.tsx
leehopper: Fixed in cfa2f1a. Verified the compiled rule in the built CSS first (`.divide-x > :not(:last-child) { border-inline-end-width: … }`), then reworded the comment to that actual mechanism — a child-scoped `:not(:last-child)` rule setting `border-inline-end-width`, which is why it is agnostic to a flex-vs-grid parent. The conclusion is unchanged; only the stated mechanism was wrong.

frontend/src/app/modules/plan/components/plan-display.helpers.ts
leehopper: Fixed in cfa2f1a — took the comment side of the fork, deliberately. The doc now says the branch takes the first segment carrying repetitions whatever its type, and the local is renamed `repeatedSegment`. I left the runtime behaviour alone: a repeated-`Recovery`-before-repeated-`Work` workout is not known to be producible, so adding a `segmentType` filter would change behaviour on the strength of a hypothetical. The two comments in the file now agree.

frontend/src/app/modules/plan/components/the-week.helpers.ts
leehopper: Fixed in cfa2f1a. Softened the claim in all three places (`isDateLogged`, the `WorkoutHeroContent` union, and `resolveHeroContent`): the shared predicate guarantees agreement only when both call sites ask about the same date, and the comments now state explicitly that an unparseable `planStartDate` or a clamped current week can leave the grid with no cell for today while the hero still reflects the real day.

frontend/src/app/modules/plan/components/plan-display.helpers.ts
leehopper: Fixed in cfa2f1a. Rewrote the header to describe the module as it now stands — the shared label maps plus the canonical UTC date pipeline and the hero summary/stat derivations. Kept the file whole rather than splitting it; the date pipeline is what the Today screen's sections share, so it belongs beside them.

leehopper:
Re: the top-level `cross-file-2` finding (PlanProjectionDto XML docs naming deleted components) — fixed in cfa2f1a. `Macro`'s summary now names `TheBlock` as its consumer, `MesoWeeks`' names `TheWeek` and `TheBlock`, and the challenger's broader catch was right: `PlanRenderingController`'s class remarks carried the same stale "macro phase strip" phrasing and were updated too.

PR #290, opened 2026-07-15
feat(split-ui): Slice 3 PR-B — Coach transcript turn kinds, timestamps, dividers, composer restyle

leehopper:
## Slice 3 (Coach) — PR-B: turn kinds, timestamps, dividers, composer restyle

The first **frontend** PR of Slice 3. Restyles the `/coach` transcript onto the Alpine turn system per `docs/specs/slice-3-coach/spec.md` §3 PR-B (DU-1/2/3/7/8). Frontend-only — no wire change (PR-A #288 already shipped the backend `loggedRun` field). Slice 3 order: **A ✅ → B (this) → C → D**.

### What's in it
- **New turn components** — `UserTurn` (D2, right-aligned bubble + `YOU · HH:MM` meta), `CoachTextTurn` (D3, no bubble, `COACH · HH:MM` clay label, inline streaming block-cursor), `DateDivider` (D7). These replace `MessageBubble`/`ChatBubble` in the transcript render path.
- **New `transcript-time.helpers`** — `formatTurnTime`, `formatDividerLabel`, `groupTurnsByLocalDay` (local wall-clock, deliberately distinct from the plan-calendar UTC pipeline) and `formatDurationSeconds`.
- **`coach-chat` rewrite** — turn-kind dispatch, local-day date-divider grouping, a single client-captured live timestamp shared by the optimistic user/coach pair, a restyled historical errored turn (copy only, **no** RETRY) vs the live retry affordance (keeps its `retryable` gate + `retry()` wiring).
- **Composer restyle** — 48px clay square send button (`ArrowUp`, `aria-label="Send message"`), new placeholder `"Ask, or describe a run to log…"`. All behavior unchanged: Enter/Shift+Enter, `initialValue`/`autoFocus`/`textareaRef`, the prefill/focus receiver.
- `CoachTextTurn` declares `loggedRun`/`units` for **PR-D** forward-compat but does not render the durable receipt yet.
- `MessageBubble` retained unmounted (the Slice-4 tool-use content-block contract), with a header note.

### Two deviations (deliberate, recorded)
1. **`liveTime` uses React's "adjust state during render" pattern, not the spec's literal `useRef`.** This repo's `eslint-plugin-react-hooks` v7 `recommended` enables `react-hooks/refs: error`, which forbids reading `ref.current` during render — the spec's snippet would fail lint. The state-pair pattern is semantically identical (one shared `HH:MM` per pending exchange, cleared on return to null) and lint-clean.
2. **The failure-surface `bg-danger-surface` token is deferred to PR-C.** The spec's PR-B text styles the retry surface with `bg-danger-surface`, but that token is defined in PR-C and the spec declares B/C independent — an ordering conflict. Resolved low-risk: PR-B stays out of `index.css`/`check-contrast` and uses `bg-secondary` + `border-l-destructive` for the retry accent, with a code comment pointing at PR-C.

### How it was built & verified
Built and adversarially verified via a sonnet multi-agent workflow: build → five review lenses (spec-compliance, correctness, conventions/a11y/tokens, test-quality, regression) → per-finding validation (refutation pass) → fix. Six findings raised, five confirmed and fixed — one real bug (**a stale live timestamp when a new message is sent directly after a prior stream errored, without passing through the null state**) plus test-coverage hardening (turn-dispatch testid scoping, `text-clay-text`/`min-h-12`/`--alp-faint` class assertions). The sixth (useRef-vs-useState) was validated as not-a-bug.

### Gates (independently re-run)
- `npm run build` ✓ · `eslint` (0 problems) ✓ · `npm run test` — **885/885** ✓ · `npm run check-contrast` — **38/38** ✓
- No backend/codegen change; no out-of-scope files touched.

### Non-negotiables (HANDOFF §8)
`role="log"`/`aria-live` transcript semantics preserved · no markdown (literal asterisks render verbatim) · icon send button has an accessible name, block-cursor `aria-hidden` · semantic tokens only, `--alp-faint` used only as decorative meta · trademark-clean.

https://claude.ai/code/session_01APgsBYCEhhvwwwndxtJhDJ

## Summary by CodeRabbit

- **New Features**
  - Added clearer transcript formatting with separate runner and coach messages.
  - Added timestamps to conversation turns and synchronized timestamps for live exchanges.
  - Grouped conversation history by local calendar day with date dividers.
  - Added streaming indicators and improved persisted error messaging.
  - Updated the composer with a send icon, accessible labeling, and refreshed placeholder text.

- **Bug Fixes**
  - Retry controls are now hidden for non-retryable errors.
  - Preserved line breaks and long-message readability in transcript content.

### DIFF
--- a/frontend/src/app/modules/coaching/components/coach-chat.component.tsx
+++ b/frontend/src/app/modules/coaching/components/coach-chat.component.tsx
@@ -1,4 +1,4 @@
-import { useCallback, type ReactElement } from 'react'
+import { Fragment, useCallback, useState, type ReactElement } from 'react'
 import { useLocation, useNavigate } from 'react-router-dom'
 import { toast } from 'sonner'
 
@@ -23,17 +23,21 @@ import {
 import { usePreferredUnits } from '~/modules/settings/hooks/use-preferred-units.hooks'
 import { AdaptationTurn } from './adaptation-turn.component'
 import { CoachComposer } from './coach-composer.component'
+import { CoachTextTurn } from './coach-text-turn.component'
+import { DateDivider } from './date-divider.component'
 import { LogConfirmationCard } from './log-confirmation-card.component'
-import { MessageBubble, type MessageRole } from './message-bubble.component'
 import { SafetyTurn } from './safety-turn.component'
+import { formatTurnTime, groupTurnsByLocalDay } from './transcript-time.helpers'
 import { TranscriptScroller } from './transcript-scroller.component'
+import { UserTurn } from './user-turn.component'
 
-// The interactive streaming-conversation panel. It unions the composed timeline
-// (interactive chat bubbles + the proactive adaptation/safety turns, reusing the
-// existing components) with the live in-flight exchange from `useCoachStream`,
-// the confirmation card, and the composer. The streamed coach reply renders as
-// plain text via the shared `MessageBubble` (`whitespace-pre-wrap`) — no
-// markdown, no raw HTML.
+// The interactive streaming-conversation panel. It unions the composed
+// timeline (turn-kind components + the proactive adaptation/safety turns,
+// reusing the existing components) with the live in-flight exchange from
+// `useCoachStream`, the confirmation card, and the composer. Persisted
+// user/coach turns render via `UserTurn`/`CoachTextTurn` (Slice 3 PR-B) —
+// both are plain-text (`whitespace-pre-wrap`), never markdown, never raw
+// HTML.
 
 const NO_TURNS: readonly ConversationTimelineTurnDto[] = []
 
@@ -60,22 +64,19 @@ const isCoachChatLocationState = (value: unknown): value is CoachChatLocationSta
   (!('focusComposer' in value) ||
     typeof (value as { focusComposer?: unknown }).focusComposer === 'boolean')
 
-interface ChatBubbleProps {
-  role: MessageRole
-  text: string
-  pending?: boolean
-}
-
-const ChatBubble = ({ role, text, pending }: ChatBubbleProps): ReactElement => (
-  <div className="flex w-full">
-    <MessageBubble role={role} content={[{ type: 'text', text }]} pending={pending} />
-  </div>
-)
-
+// A persisted historical errored turn has no live `retry()` to call — RETRY
+// belongs only to the live-stream `RetryAffordance` below. Restyled as a
+// `CoachTextTurn`-shaped block (mono `COACH` label, plain body) so it reads
+// as part of the transcript rather than a bare disconnected line.
 const ErroredCoachNote = (): ReactElement => (
-  <p data-testid="coach-errored-turn" className="px-1 text-sm text-muted-foreground">
-    That reply didn&apos;t go through.
-  </p>
+  <div data-testid="coach-errored-turn" className="flex flex-col gap-1">
+    <span className="font-mono text-[10px] font-semibold tracking-[0.1em] text-clay-text">
+      COACH
+    </span>
+    <p className="font-body text-[14.5px] leading-[1.55] text-muted-foreground">
+      That reply didn&apos;t go through.
+    </p>
+  </div>
 )
 
 const SafetyNotice = ({ notice }: { notice: CoachSafetyNotice }): ReactElement => (
@@ -99,11 +100,18 @@ const RetryAffordance = ({ error, onRetry }: RetryAffordanceProps): ReactElement
   <div
     role="alert"
     data-testid="coach-error"
-    className="flex items-center justify-between gap-3 rounded-md border border-border bg-secondary px-3 py-2 text-sm"
+    // Failure-surface token bg-danger-surface lands in PR-C (spec §3 PR-C); using bg-secondary here until then.
+    className="flex items-center justify-between gap-3 rounded-md border-l-[3px] border-l-destructive bg-secondary px-3 py-2"
   >
-    <span className="text-destructive">{error.message}</span>
+    <span className="font-mono text-[12px] text-foreground">{error.message}</span>
     {error.retryable && (
-      <Button type="button" variant="outline" size="sm" onClick={onRetry}>
+      <Button
+        type="button"
+        variant="outline"
+        size="sm"
+        onClick={onRetry}
+        className="font-condensed text-[11px] font-semibold tracking-[0.08em] uppercase"
+      >
         Retry
       </Button>
     )}
@@ -120,9 +128,18 @@ const TimelineRow = ({
   // Narrow on the payload null-ness — exactly one of interactive/proactive is set.
   if (turn.interactive !== null) {
     if (turn.interactive.isErrored) return <ErroredCoachNote />
-    const role: MessageRole =
-      turn.kind === CONVERSATION_TIMELINE_TURN_KIND.user ? 'user' : 'assistant'
-    return <ChatBubble role={role} text={turn.interactive.content} />
+    const time = formatTurnTime(turn.createdAt)
+    if (turn.kind === CONVERSATION_TIMELINE_TURN_KIND.user) {
+      return <UserTurn content={turn.interactive.content} time={time} />
+    }
+    return (
+      <CoachTextTurn
+        content={turn.interactive.content}
+        time={time}
+        loggedRun={turn.interactive.loggedRun}
+        units={units}
+      />
+    )
   }
   return turn.proactive.role === CONVERSATION_ROLE.systemSafety ? (
     <SafetyTurn turn={turn.proactive} />
@@ -150,6 +167,27 @@ export const CoachChat = (): ReactElement => {
   } = useCoachStream()
   const [confirmLog, { isLoading: isConfirming }] = useConfirmConversationalLogMutation()
   const navigate = useNavigate()
+
+  // Both live rows (the optimistic user bubble, the streaming coach block)
+  // share one client-captured `HH:MM` so the pair doesn't visibly disagree
+  // as tokens arrive. Captured once per exchange via React's blessed
+  // "adjust state during render" pattern (comparing against a snapshot of
+  // the previous render's `pendingUserMessage`, tracked in state rather
+  // than a ref — `react-hooks/refs` forbids reading `ref.current` during
+  // render): on ANY transition of `pendingUserMessage` to a new value,
+  // `liveTime` is recomputed — non-null yields a fresh `new Date()` capture,
+  // null clears it. This covers not just null -> non-null but also a direct
+  // non-null -> different-non-null transition (e.g. sending a new message
+  // after a prior exchange errored without going through null), which would
+  // otherwise leave a stale timestamp on the new optimistic bubble. This
+  // `new Date()` read is the ONE sanctioned live wall-clock read — every
+  // persisted turn always uses its server `createdAt` instead.
+  const [prevPendingUserMessage, setPrevPendingUserMessage] = useState(pendingUserMessage)
+  const [liveTime, setLiveTime] = useState('')
+  if (pendingUserMessage !== prevPendingUserMessage) {
+    setPrevPendingUserMessage(pendingUserMessage)
+    setLiveTime(pendingUserMessage !== null ? formatTurnTime(new Date().toISOString()) : '')
+  }
 
   const handleConfirm = useCallback(async (): Promise<void> => {
     if (card === null) return
@@ -182,6 +220,11 @@ export const CoachChat = (): ReactElement => {
   // streaming text length in keeps the live partial pinned to the bottom.
   const scrollKey = timeline.length + (pendingUserMessage !== null ? 1 : 0) + streamingText.length
 
+  // Date dividers are computed over the persisted timeline turns only — the
+  // live pending/streaming exchange belongs to "today" and renders after
+  // the last group with no live-introduced divider (§4.1/§6).
+  const dayGroups = groupTurnsByLocalDay(timeline)
+
   return (
     <section
       aria-labelledby="coach-chat-heading"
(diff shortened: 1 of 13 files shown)


frontend/src/app/modules/coaching/components/coach-chat.component.tsx
```diff
@@ -151,6 +168,27 @@ export const CoachChat = (): ReactElement => {
...
+  // `liveTime` is recomputed — non-null yields a fresh `new Date()` capture,
+  // null clears it. This covers not just null -> non-null but also a direct
+  // non-null -> different-non-null transition (e.g. sending a new message
+  // after a prior exchange errored without going through null), which would
+  // otherwise leave a stale timestamp on the new optimistic bubble. This
+  // `new Date()` read is the ONE sanctioned live wall-clock read — every
+  // persisted turn always uses its server `createdAt` instead.
+  const [prevPendingUserMessage, setPrevPendingUserMessage] = useState(pendingUserMessage)
+  const [liveTime, setLiveTime] = useState('')
+  if (pendingUserMessage !== prevPendingUserMessage) {
```
leehopper: **💡 [LOW] liveTime isn't refreshed when RETRY re-sends the identical message (stale timestamp)**

The `liveTime` guard recomputes only when `pendingUserMessage` changes by **string value**. But in `use-coach-stream.hooks.ts` the `error` reducer branch leaves `pendingUserMessage` unchanged (the bubble stays beside the retry affordance), and `retry()` re-sends `lastMessageRef.current` — the *same* text — dispatching `start` with that identical string. So the `X → X` transition is invisible to this guard: the `if` never runs, and both the retried optimistic `UserTurn` and the streaming `CoachTextTurn` render the timestamp captured at the **original failed attempt** (stale by however long the user waited before clicking RETRY).

The block's own comment claims to prevent exactly this staleness, but only covers the `non-null → different-non-null` case (a *new* message) — not the identical-text retry, which is the primary RETRY path.

**Fix:** stamp `liveTime` off the attempt/exchange identity rather than the message text — e.g. recompute it in `send`/`retry` (or key it off the fresh `clientMessageId`), or reset it whenever an exchange (re)starts regardless of whether the text changed. Confirmed by 3 independent agents and a blind challenger.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
```diff
@@ -151,6 +168,27 @@ export const CoachChat = (): ReactElement => {
...
+  // as tokens arrive. Captured once per exchange via React's blessed
+  // "adjust state during render" pattern (comparing against a snapshot of
+  // the previous render's `pendingUserMessage`, tracked in state rather
+  // than a ref — `react-hooks/refs` forbids reading `ref.current` during
+  // render): on ANY transition of `pendingUserMessage` to a new value,
+  // `liveTime` is recomputed — non-null yields a fresh `new Date()` capture,
+  // null clears it. This covers not just null -> non-null but also a direct
+  // non-null -> different-non-null transition (e.g. sending a new message
+  // after a prior exchange errored without going through null), which would
+  // otherwise leave a stale timestamp on the new optimistic bubble. This
```
leehopper: **💡 [LOW] No test for the documented non-null → different-non-null liveTime transition**

This comment documents a deliberate case — a *new* message sent right after a prior error, without passing through null — as the reason the state-pair pattern exists. But no test in `coach-chat.component.spec.tsx` drives two different non-null `pendingUserMessage` values across a rerender; the only `liveTime` test covers a `null → non-null` fresh send. A regression in the guard would go uncaught. Related to the `liveTime` staleness bug above — a good fix adds tests for **both** the different-text transition and the identical-text retry.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
```diff
@@ -182,6 +220,11 @@ export const CoachChat = (): ReactElement => {
   // streaming text length in keeps the live partial pinned to the bottom.
   const scrollKey = timeline.length + (pendingUserMessage !== null ? 1 : 0) + streamingText.length
 
+  // Date dividers are computed over the persisted timeline turns only — the
+  // live pending/streaming exchange belongs to "today" and renders after
+  // the last group with no live-introduced divider (§4.1/§6).
+  const dayGroups = groupTurnsByLocalDay(timeline)
```
leehopper, line 226: **💡 [LOW] Date-divider grouping untested with proactive (adaptation/safety) turns**

`groupTurnsByLocalDay(timeline)` runs over the full timeline — interactive **and** proactive turns. But the divider-placement test uses interactive-only (`turnAt`) fixtures, and the one mixed-kind test keeps every turn on the same day and never asserts on `date-divider`. No test covers a day boundary that includes a proactive turn, or that a proactive turn buckets into the same local-day group as neighbouring interactive turns. The implementation is kind-agnostic and likely correct — this is a coverage gap, not a defect.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
```diff
@@ -195,11 +238,18 @@ export const CoachChat = (): ReactElement => {
         turnCount={scrollKey}
         className="min-h-0 flex-1 rounded-md border border-border bg-card p-4"
       >
-        {timeline.map((turn) => (
-          <TimelineRow key={turn.turnId} turn={turn} units={units} />
+        {dayGroups.map((group) => (
+          <Fragment key={`day-${group.turns[0]?.turnId ?? group.label}`}>
```
leehopper: **💡 [LOW] Unreachable, unsafe-if-reached fallback in the day-group Fragment key**

`groupTurnsByLocalDay` never emits an empty group (every group is seeded `turns: [turn]` and only appended to), and this repo has no `noUncheckedIndexedAccess`, so both the `?.` and the `?? group.label` branch are dead code. And `group.label` has no year, so it isn't globally unique (the same weekday/month/day a year apart collides) — the fallback would produce duplicate React keys *if* it ever ran. Simplify to key directly off the (always-present) first turn's id:

```suggestion
          <Fragment key={`day-${group.turns[0].turnId}`}>
```

frontend/src/app/modules/coaching/components/coach-composer.component.tsx
```diff
@@ -69,21 +70,27 @@ export const CoachComposer = ({
...
     >
       <Textarea
         ref={textareaRef}
         aria-label="Message your coach"
-        placeholder="Ask a question or describe a run to log…"
+        placeholder="Ask, or describe a run to log…"
         value={value}
         onChange={(event) => setValue(event.target.value)}
         onKeyDown={handleKeyDown}
         rows={2}
```
leehopper: **💡 [LOW] Composer textarea keeps rows={2} instead of the spec-locked rows={1}**

Spec §3 PR-B (and DU-7) lock the composer to `<Textarea rows={1}>` with `min-h-12` for a **48px resting height** paired with the 48px clay send square. This PR restyled this exact element (added the new placeholder + `min-h-12`) but left `rows={2}`. `min-h-12` is a floor, not a cap: at `text-sm` line-height + `py-2`, a 2-row textarea intrinsically computes ≈56–58px, so it rests taller than the 48px target. There's no auto-grow logic to pull it back down.

```suggestion
        rows={1}
```

frontend/src/app/modules/coaching/components/coach-text-turn.component.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { cn } from '@/lib/utils'
+import type { PreferredUnits } from '~/api/generated'
+import type { LoggedRunSummaryDto } from '~/modules/coaching/models/conversation.model'
+
+export interface CoachTextTurnProps {
+  content: string
+  time: string
+  /** Appends the clay block-cursor inline at the end of the body when a live stream is still in flight. */
+  streaming?: boolean
+  /** PR-D wires this into a durable `LoggedRunReceipt`; left undefined by every PR-B call site. */
```
leehopper: **💡 [LOW] JSDoc embeds planning-phase "PR-D" forward references + a literal future-code snippet**

Root `REVIEW.md` § Code comments bans planning-phase forward references (only Spec / DEC / R-NNN symbolic refs allowed) and TODO/FIXME. This JSDoc references "PR-D" repeatedly and embeds `// PR-D: render <LoggedRunReceipt summary={loggedRun} units={units}> beneath the body.` — functionally a TODO. Judgment call: these are deliberate forward-compat notes, but per the literal rule they should be rephrased as evergreen "why" comments (or cite a DEC/spec section) before merge.

frontend/src/app/modules/coaching/components/transcript-time.helpers.ts
```diff
@@ -0,0 +1,114 @@
+// Transcript timestamp + local-day-grouping helpers (spec §4.1) — LOCAL
+// wall-clock, deliberately distinct from `plan-display.helpers`' UTC
+// plan-calendar pipeline (§4.2). A message shows the time it was sent in
+// the READER's timezone, not the server's or the plan's anchor timezone, so
+// every getter here is a local (`getHours`/`getDay`/`getMonth`/`getDate`)
+// read — never a `getUTC*` one.
+//
+// `formatDurationSeconds` (§4.3) lives here too: it is the receipt
```
leehopper: **💡 [LOW] File-header comment narrates "PR-D scope" and a two-PR split**

Same `REVIEW.md` § Code comments rule. The `§4.3` spec citation is allowed, but "the receipt itself (`LoggedRunReceipt`) is PR-D scope — shipping the helper … avoids a same-file split across two PRs" is PR-sequencing narrative — analogous to the banned "Slice 1 only" pattern. Trim to the load-bearing invariant (why `formatDurationSeconds` lives here) without the PR-sequencing rationale.

frontend/src/app/modules/coaching/components/message-bubble.component.tsx
```diff
@@ -1,5 +1,11 @@
 import type { ReactElement } from 'react'
 
+// Retained, unmounted-by-transcript as of Slice 3 PR-B: `coach-chat.
+// component.tsx`'s transcript now renders `UserTurn`/`CoachTextTurn`
+// instead of `MessageBubble`/`ChatBubble`. This component (and its
+// content-block filter) stays because it is the documented Slice-4
```
leehopper: **💡 [LOW] Retained-component comment cites a not-yet-built "Slice-4" contract**

Same `REVIEW.md` § Code comments rule — a bare slice-number forward reference, structurally like the banned "Unit 5 will". The "why kept, not dead code" intent is legitimate and worth preserving; prefer citing a **DEC number** over a slice label. Note "Slice 4" is ambiguous — it means different things in the MVP-0 vs SPLIT-UI cycles.

leehopper (commented):
## Deep Review — PR #290 (Slice 3 PR-B)

Multi-agent review (7 concern agents → factual verification → Sonnet validation → blind challenge). 14 findings raised → **8 delivered**, all blind-challenge-confirmed. **No security, correctness-critical, or high-severity issues — all findings are LOW severity.** 2 main findings (a `liveTime` staleness bug on identical-message retry, and a `rows={2}` deviation from the spec-locked `rows={1}`), plus 6 improvement suggestions (test-coverage gaps, comment-convention nits, a dead-code key fallback).

Verified clean: XSS/markdown-injection surface (all content is escaped JSX text, no `dangerouslySetInnerHTML`), cross-file impact (no dangling `MessageBubble` consumers), type design, trademark, semantic tokens.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
leehopper: Fixed — `liveTime` now keys off `isStreaming`'s false -> true transition instead of `pendingUserMessage` string equality, so a RETRY that resends identical text gets a fresh capture instead of inheriting the failed attempt's stale timestamp. See 32bffae8.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
leehopper: Added coverage for both the identical-text-retry and the different-text `liveTime` transitions in `coach-chat.component.spec.tsx`.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
leehopper: Added a test covering a day boundary where the second group mixes a proactive (restructure) turn with an interactive turn.

frontend/src/app/modules/coaching/components/coach-chat.component.tsx
leehopper: Simplified to key directly off `group.turns[0].turnId` as suggested — the fallback was dead code.

frontend/src/app/modules/coaching/components/coach-composer.component.tsx
leehopper: Fixed — `rows={1}` now matches the spec-locked 48px resting height.

frontend/src/app/modules/coaching/components/coach-text-turn.component.tsx
leehopper: Rephrased to cite spec §3 PR-D instead of a bare 'PR-D' forward reference, and dropped the embedded future-code snippet.

frontend/src/app/modules/coaching/components/transcript-time.helpers.ts
leehopper: Trimmed the PR-sequencing narrative from the file header, per the same rule CodeRabbit flagged on this file.

frontend/src/app/modules/coaching/components/message-bubble.component.tsx
leehopper: Rephrased per CodeRabbit's parallel finding on this file — dropped the Slice-4/coach-chat.component.tsx references, kept the 'why kept, not dead code' invariant.

